PR #4, opened 2022-11-19
Final Lit Review Uploaded

stephankokkas:
This is my final upload for the lit review doc I've been working on. If you could review for me and approve if happy :)


akudilcz (changes requested):
Hi @stephankokkas This literature review is a really amazing piece of work for such a short time period!  Well done!  My main comment is that although you have a list of references, these are not referenced by the main body of text - it would be good to understand which reference contributed to each of the paragraphs and, in particular, where did those tables for comparison come from?  Can we settle on using standard Deakin referencing style as per [Deakin guide to referencing](https://www.deakin.edu.au/students/studying/study-support/referencing) ?

stephankokkas:
Done - ill make the changes by EOD

stephankokkas:
updated to include referencing

akudilcz (approved):

PR #12, opened 2022-11-28
Sk data eng pipeline 

stephankokkas:
tested and working on both Mac and windows - please ensure you follow the README file closely as it is important you have all the required libraries installed. 

Thanks


stephankokkas:
The final pipeline will load all the final tensors into memory and prep them for training - any pipeline in-between will not as they involve raw mp3 files.

stephankokkas:
If you have any issues running the notebook please let me know and I will resolve. It is very challenging writing code that works with all paths for os' but it works for me on both windows and Mac

akudilcz:
In order to support windows users, we need to force:

tensorflow==2.10.1
tensorflow-gpu==2.10.1

Then need to fix compiler issues.  Currently I can't run the script on windows.

akudilcz (changes requested):
Currently I can't run the script on windows.

In order to support windows users, we need to force:
tensorflow==2.10.1
tensorflow-gpu==2.10.1

Need to add these library dependencies to env_dev.txt: 
tensorflow_io
pydub

Then need to fix compiler issues.

stephankokkas:
Thanks for your review - I am logging off for today but I will work on it first thing in the AM.

akudilcz:
I managed to get it to run, but I am having trouble understand how to load the data into a data pipeline.  

I was expecting the following output style:

output/
  **train/**
     class_1/
         class_1_sample1.pt
         class_1_sample2.pt
     class_2/
         class_2_sample1.pt
         class_2_sample2.pt

output/
  **test/**
     class_1/
         class_1_sample1.pt
         class_1_sample2.pt
     class_2/
         class_2_sample1.pt
         class_2_sample2.pt

and so on... similar to image file format, but with tensors instead of images.

the tensor files look good - but they are not in the above layout format so are hard to load - would it be possible to organise them in that above format?

Happy to have a chat tomorrow to find out what your thoughts are on how to load these :)

stephankokkas:
Hey @akudilcz, the file pipeline handles the loading of the files into datatframe but happy to sort the output into a more structured file system. What did you have to change so that the pipeline worked?

akudilcz:
I also looked at the .csv files and they appear to be broken which explains why they are so small -> the actual melspectrogram data isn't in there only the first few entries and last few with dots '. . . ' in between!  Take a close look at the first entry - copy and paste it into another editor and you'll see what I mean :)

stephankokkas:
Hey Andrew, yeah the TO_CSV function was only for debugging - if you inspect the frames when they are loaded in memory they contain the complete tensors. Are exporting the frames required?

akudilcz:
> Hey Andrew, yeah the TO_CSV function was only for debugging - if you inspect the frames when they are loaded in memory they contain the complete tensors. Are exporting the frames required?

That makes sense.  I don't think we need the frames.   Just the organised folders with the data files would work fine! :)

stephankokkas:
Final pipeline fixed and working. Massive thank you to @akudilcz for your review, response, feedback, and suggestions

PR #14, opened 2022-11-30
Sk   tf.io fixes on data pipeline

stephankokkas:
Added logic to allow export and read using tf.io to avoid issue with model training. Thanks @akudilcz for the support!


akudilcz (approved):
seems to work much better!

PR #15, opened 2022-11-30
Ak baseline model htsat

akudilcz:


stephankokkas (approved):
+3,213! Wow, great effort and awesome contribution. Model looks super complex and really interesting. LGTM!

PR #18, opened 2022-12-05
Ak design update week 4

akudilcz:


crocodillian:
Looks good, nice update to the design.

crocodillian (approved):
Great work Andrew! Can eventually expand on the simulator and the front end devices but the concept is there.

PR #19, opened 2022-12-05
Ak baseline cnn model

akudilcz:


crocodillian:
Looks good, 85% is very impressive at this stage!

crocodillian (approved):
The baseline model is excellent! Good work on your implementation and the tensorboard visualizations are really fascinating! Now to train it on another dataset.

PR #34, opened 2022-12-11
retrained tfimm model and eda notebook fix

akudilcz:


akudilcz:
Fixed eda notebook too.  It didn't run out of the box.

PR #60, opened 2023-03-24
Sk ak (sim) design+requirements

stephankokkas:
We now have a POC of a working simulator which can generate an otways map, populate it with microphones, simulate an animal event, and triangulate its event. It then prints out a json dump of the event lat and long values with a few ID's. The algorithms work well but Andrew and I will now work on a new triangulation algorithm so that we can triangulate an event based only on the time a microphone was triggered at.


akudilcz (approved):
Looks fine - In the future we may consider have a single env - rather than one for each component - but the flexibility at this early stage is a good idea.

PR #75, opened 2023-04-19
Merge database and API prototypes

tzhai92:
Included: 

- Sample data and code used to generate it
- Prototype API built using FastAPI
- Tests for using MQTT
- Tests on database connection and querying


src/Prototypes/data/database/.env
akudilcz, line 1: I can see password here.  Is that an issue?

tzhai92:
Added gitignore for environment secrets

src/Prototypes/data/database/.env
tzhai92: updated gitignore so now you wont see passwords hopefully

akudilcz (approved):

PR #82, opened 2023-04-22
Hmi dev 2

Q-Cooke:
Releasing current UI prototype from HMI dev to main so we can make a start on integration work with the API


stephankokkas (approved):
Approved per request

stephankokkas:
Approving as code has already passed review.

PR #83, opened 2023-04-22
Tz api

tzhai92:
I've fixed some more bugs with the API, including location errors and timezone issues. Also fixed the .gitignore situation so running anything connecting to the database will need a .env file from me to protect credentials.


akudilcz (approved):
Changes look good

PR #96, opened 2023-04-30
Sk database containerisation

stephankokkas:
src/Components/MongoDB/test_connection.py is a file which tests the connection strings.


stephankokkas:
forward merged into ak engine and store integration. Will only merge one branch.

PR #104, opened 2023-05-05
Applying autoencoder to existing model in optimised pipeline

Ank9812:


akudilcz (changes requested):
The way the model is built so far once the 'encoder' model part is incorporated into the final model, the weights are not frozen - this means the encoder part just becomes the front end CNN for the model and all the weights are re-trained in the 2nd fit process.  Please update the code to freeze the weights on the encoder model before running the 2nd fit process.   You also need to increase the number of filters used - the small numbers mean the encoder is not able to encode complex image data.

akudilcz (approved):

PR #105, opened 2023-05-06
Create multilabel_engine_pipeline.ipynb

RyanBowenDeakin:
A starting point for a new pipeline with multi-label classification, this is just small tweaks to Andrew's work and hasn't been tested yet, just a place to start working from


akudilcz:
Hi @RyanBowenDeakin, where you have these lines of code in the model fit cell:

possibleLabels = 5 #Not sure yet how to determine this value properly
model.add(tf.keras.layers.Dense(possibleLabels, activation="sigmoid"))

I don't think they are actually doing anything - can you confirm whether they are needed?  They seem to be called after the model has already been compiled - and the model structure is not affected (see the printout of the model).

This is a multilabel classification - so I can see the outputs are now following a binomial distribution, however the input pipeline is not doing any mixing of species audio - I couldn't see where there you would get sample labels with more than one species - in which case this is falling back to single label classification.

akudilcz (approved):
Approved on the basis, the is a starting point for implementation.

PR #128, opened 2023-07-20
added the first file to try

mic-0-hal:
added a folder where students can make their first git commit into


mic-0-hal:
this contains the first notebook file for the juniors to use

mic-0-hal:
please merge into main.

PR #137, opened 2023-07-22
changed jupyter file

mic-0-hal:
was working in the wrong branch, and i renamed and modified the file i was working on. I have since duplicated it to see if that pulls through and deletes the other file


Q-Cooke (approved):
Looks good to me.

PR #139, opened 2023-07-22
cleaned up install files

mic-0-hal:
removed duplicate install file which was named incorrectly


Q-Cooke (approved):
Good to merge.  I have spoken.

PR #146, opened 2023-07-24
completed task and added audio files 

mic-0-hal:
also added audio files to directory as they were missing


Q-Cooke (approved):
LGTM cheers

PR #148, opened 2023-07-24
Create introduction_to_audio_-RL-checkpoint.ipynb

bobjoneswofriley:
Individual Dev Environment Onboarding first Pull request test.


Q-Cooke:
Your branch name is not following the convention 100% correctly, it should be: RL/intro_to_audio

Q-Cooke (approved):
You have completed this task efficiently, which I can respect.  However If you took advantage of the opportunity to practice a bit more Python, you would generate more evidence that you can reference for your on-track tasks later.  Just something to consider.

PR #151, opened 2023-07-25
Tn/enable authentication

VietTung:
Set up authentication and authorization for HMI component using JWT:

- added **model** directory for Roles (admin/user) and Users collections design in mongoDB
- added **config** directory for environment setting
- added **controller** and **route** directories to handle user authentication and user role determination
- added **middleware** directory to handle user verifiication using JWT token
- added **login.html** , **auth-form.css** file for default login page
- modified **server.js** to consume aforementioned components
- added mongo-express container to monitor echo store container

**please run npm install inside Components/HMI/ui directory as new packages are added in**


Q-Cooke:
This has been merged into HMI-dev-10.  Thanks heaps Tung, excellent work

PR #157, opened 2023-07-27
Changes and Addition to species - Yohjit Chopra

Yohjit-2910:
I have made changes and added important information about the species that were assigned to me (Yohjit).


src/Components/Simulator/requirements.txt
mic-0-hal, line 1: checking these requirements are to be removed?

Q-Cooke (approved):
Excellent work Yohjit, thank you

PR #162, opened 2023-07-30
Intro to audio

cpavia:
Using new naming convention


src/Environments/Install_ EchoNet_README.md
mic-0-hal, line 1: looks like this got pulled through in the branch. no changes to the file though

mic-0-hal (changes requested):
there is a file in .ipynp_checkpoints that still has my initial, and a few more with my initial, but blank. and then yours which has your changes.

Also a readme that you may have used?

PR #181, opened 2023-08-08
Ee/cp/add event detection to live audio pipeline

cpavia:
I have added real-time event detection using the pyaudio library to capture live audio from the microphone and analyse the audio data to identify events or patterns. The detect_events function monitors the audio data and triggers an alert when the audio exceeds a predefined threshold. This enables the user to be promptly notified of detected events while the audio stream is being monitored.


Q-Cooke (approved):
LGTM, thanks Charmaine.

PR #182, opened 2023-08-11
Ee/Jv/Adding-Event-Detection-to-live-audio-pipeline

Joel213141:
- I have added a Spectral Centroid based event detector to the real time event detection pipeline.
- Corrected code to convert the audio data from bytes to 16-bit integer values for event detection. 
- Added additional code to convert audio data to the float format required to be fed into the 'Librosa' module functions. 

Spectral Centroid appear to be more robust than amplitude and zero-crossing rate thresholding after testing the pipeline against various types of audio input data.


Q-Cooke (approved):
LGTM, thanks Joel

PR #184, opened 2023-08-13
PANNs_errors_Test

Jeffchenchen:
This is a Trail due to still errors when run


mic-0-hal (commented):
How are you going with this task?
Do you have a compiled model?

PR #186, opened 2023-08-13
notebook to implement augmentations to increase dataset variety

abhitgahlot:
notebook to implement augmentations to increase dataset variety and updated pipeline to add some more augmentations like background noises and short noises

### DIFF
--- a/src/Prototypes/engine/audio_augmentations.ipynb
+++ b/src/Prototypes/engine/audio_augmentations.ipynb
@@ -1,0 +1,134 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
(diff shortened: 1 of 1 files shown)


Q-Cooke (changes requested):
please use EE instead of engine in your branch name mate in future, please acknowledge you will do this with a comment before I merge this

abhitgahlot:
I'll make sure to use EE instead of engine for the branch name in future.

mic-0-hal (changes requested):
Can you run the pipeline with these augmentations please.
And add a notebook to this branch describing what you have done with some stats/figures/etc. of the benefit to these augmentations

PR #188, opened 2023-08-13
Sb/apiauthentication

CheanseanBo:
API Authentication: signin, signup functionalities tested and working.
signout may need more modification for clearing the correct user session


Q-Cooke (approved):
Looks good Sean, great work.

PR #194, opened 2023-08-19
EHMI/SB/Expansion-User-Schema

CheanseanBo:
Tested and working.
New change:
API expansion to user schema adding DoB, Address, Phone Number and Organization. 
Awaiting SignUp page to be created to add new field.
Note for those creating signup page. Refer to this task to see what schema field are required.
Components/API/schemas.py


Q-Cooke:
This is already in HMI-dev-10

PR #205, opened 2023-08-23
testing incorrect renaming and local model and named branch incorrectly

mic-0-hal:
Created a local model and successfully ran it in local echonet. This branch is named incorrectly to test what happens when renaming is required.


Q-Cooke (changes requested):
Please change branch name mate

PR #215, opened 2023-08-29
Librosa pipeline

dlees27:
Added a pipeline that uses librosa to convert audio to a mel spectrogram


cpavia:
Hi, can you please use the correct naming convention. It can be found in the main chat of Project Echo. Thanks!

PR #218, opened 2023-08-30
EE/SW/Transfer learning using OpenL3 model

939781369:
Transfer learning using OpenL3 model


cpavia:
Hi Sizhe, can you please review:

- Need to import numpy library to ensure 'create_subset' can run
- Suggest to have all necessary libraries at the top of the script instead of in each code cell 
- receiving 'data' error in 'preprocess_audio_data' cell
- receiving error in line 7 due to incorrect naming of 'preprocessed_data' function

Can you fix these and I'll check it over again? Thanks!

mic-0-hal (changes requested):
Hi,
The accuracy of this model is low, and I dont think it will be improved if I run it. 
The optimised pipeline needs to use this transfer model. 
Please make a copy of the optimised pipeline and  incorporate this into that file.

Save it in the "Transfer Learning Tasks" folder in prototypes

mic-0-hal (changes requested):
Hi Siheng, I cannot see where you have run this file with OpenL3 as the base model.

The code added at the end doesn't relate to the pipeline.
In the master version under build model. Line 11. there is an imported model from imagenet. This is function builds the model to be used by the pipeline. It is here that the new model using OpenL3 is to be incorporated. 
The first step is to get the model working in the pipeline file.
Then to train the model and test its performance

mic-0-hal:
You have made good progress with this. I am unable to run it locally, but will commit this work so far. It can be picked up on next trimester.
It seems so close to working and will be great to test it as model to adopt later

mic-0-hal (approved):

PR #219, opened 2023-08-30
EE/SW/Create README.md

939781369:
insect data README.md file


mic-0-hal (approved):
approved. this directory looks ok, but when we get more files coming in we may decide to change it.
all good for now though

PR #222, opened 2023-08-31
Updated-on-Signup-Form

CheanseanBo:
Created some more schema to user signup. Will match with the API user schema


RyanBowenDeakin (changes requested):
I'd suggest some of these should be dropdowns rather than text fields?
Roles and Gender have a name of 'confirmpassword', should be unique.

PR #223, opened 2023-09-01
EE/MH/created-folders-for-each-prototype-type

mic-0-hal:
Added folders to prototypes/engine to keep the different tasks organised


cpavia (changes requested):
Hi Michael, can you review spelling error on line 7, otherwise all good. Cheers

mic-0-hal (commented):
corrected spelling as suggested

cpavia (approved):

PR #227, opened 2023-09-02
EE/CP/Added-keras-version-instructions

cpavia:
Added a note about the Keras version to the optimised engine pipeline file in Components. Also updated the master version in the Prototype folder for other engine students.


Q-Cooke (approved):
LGTM, Thanks Charmaine

PR #228, opened 2023-09-02
EE/CP/event-detection-standalone

cpavia:
Created a standalone file with event detection. Tested and ready to deploy into a live testing environment.


Q-Cooke (approved):
LGTM

PR #229, opened 2023-09-02
Librosa Pipeline

dlees27:
Added a basic pipeline that utilises Librosa


cpavia:
Hi, can you please use the correct naming convention. It can be found in the main chat of Project Echo. Thanks!

PR #232, opened 2023-09-04
Ee/cg/transfer learning vggish

chris-gao-36:


mic-0-hal (changes requested):
This request has changes in Components as well as prototypes. Also have changes in the SW Generating heat map file.

There is no VGGish model in the files.

Also, make sure the model compiles without errors.

PR #233, opened 2023-09-04
EE/MH/created-folders-for-each-prototype-type

mic-0-hal:
added new folders


Q-Cooke (approved):
LGTM Mate, thanks heaps

PR #246, opened 2023-09-08
EE/MH/create-python-module-library

mic-0-hal:
added a template workbook that has a function to display mel spectrograms. This will be added to the


mic-0-hal:
I have moved the location of the spec display function workbook and created a folder in prototypes for it

Q-Cooke (approved):
LGTM

PR #248, opened 2023-09-09
EE/DLe/Librosa-Pipeline

dlees27:
A Pipeline that uses librosa instead of Matplot lib

### DIFF
--- a/src/Prototypes/engine/pipline_librosa.py
+++ b/src/Prototypes/engine/pipline_librosa.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Aug 28 13:39:33 2023
+
+@author: User
+"""
+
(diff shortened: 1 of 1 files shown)


mic-0-hal (changes requested):
Can you pease add a readme file to accompany this and markup the code with some more detail.
This task doesn't require trimming the audio, but displaying the audio file using librosa.
Start with the notebook I put in the trello card and work from that to grow the function in there.

This function call will be used throughout all project echo when plotting so I will need to take as many parameters as you can think of and display the plot accordingly

dlees27:
Thanks for the Feedback. I will work on this tomorrow, and submit anouther pull request by EOD

dlees27:
I am deleting this pull request as I have actioned the requested changes under EE/dle/librosa img display

PR #249, opened 2023-09-09
EE/SW/Extracting-Animal-and-Bird-Sounds

939781369:
YAMNet model


mic-0-hal (changes requested):
Hi Siheng,
The task of detecting events using yamnet is already what Rohit has created. 
Charmaine, Joel, Sizhe have also made progress on this role.
This notebook hasn't been applied to any of our training data.
I would talk to Rohit, Charmaine, Joel, Sizhe about the work they have done.

What is required next is to detect bird sounds in our training files and extract 10 second (time length yet to be confirmed) audio when detected. Note from the file I posted in the group chat that single files can be minutes (up to 17 mins) long and we need to extract the segments from our training files.

Also, the output should be more than just a .csv.
Create a well formatted pandas dataframe. And add visualisations where possible. 

Also, when creating a branch, use "EE/SWe/" in the branch name as requested, to distinguish your branches from Sizhe's

mic-0-hal:
You have also made this branch straight off of main. Files need to be added to the prototypes folder, and there are subfolders in there to put your work, based on the type

PR #250, opened 2023-09-09
EAPI/SB/cred_change

CheanseanBo:
Reviewed and tested. 
New functionalities that can change user credential such as gender, DoB etc..
Changes made to an account requires jwtToken upon logging in.
Signout functionalities has been created as well, store logout_token inside the client side.


Q-Cooke (approved):
LGTM

PR #251, opened 2023-09-09
EE/CP/Remove-files

cpavia:
Removed original event detection and segmentation file as these two components are now saved separately


Q-Cooke (approved):
Thanks Charmaine

PR #255, opened 2023-09-09
EAPI/SB/Delete_Account

CheanseanBo:
Delete Account functionality.
Converted the previous token from using userId to ObjectId of _id instead.


Q-Cooke (approved):
LGMT, thanks mate

PR #260, opened 2023-09-10
Ee/d le/librosa img display

dlees27:
Added a script that includesw a function to display a wave funciton, and anouther fuction to show a mel spectrogram.


mic-0-hal (changes requested):
this is a good start.
eventually we will have this as a script, but for now can you make a workbook for this. Add a title and description of this function and how it will be used in markdown.
The readme descriptions can do as comments in the cells with the functions.

Then run the notebook with some data to display it functioning and so we can test it too.

It will need to take in many more variables as described to be able to display this as image shapes, grid shapes, image labels, etc

dlees27:
I have moved the script over to Jupyter notebook and added functionality to display the shape of the mel-spec, and allow the user to specify the size of the image

mic-0-hal:
I am working on this in between other priority tasks, will get back to you soon

mic-0-hal (changes requested):
Hi Dane, there is a checkpoints folder in Components/Engine.
Please add a new folder in Prototypes/engine/ with for your task.
Otherwise can be merged

Q-Cooke (approved):
LGTM thanks heaps

PR #261, opened 2023-09-10
Ee/AKk/kaggle birdclef2023 identifying aus birds

AtharvaKulkarni26:


mic-0-hal (changes requested):
Can you put these files in folders based on species name, and then share this link?
Confirming there were only 20 Aus bird in Birdclef? 

The notebook can have more detail please. Use markdown sections to report what has been done, and the data frames can be tidied up and presented better

mic-0-hal:
Have you found more Aus bird species elsewhere?

Birdclef data is sourced from https://xeno-canto.org/explore?query=australia and there are plenty of Aus species there to source

mic-0-hal:
Make another pull request with these changes please

PR #262, opened 2023-09-10
EE/MH/convert-2-wav-pipeline

mic-0-hal:
Created a notebook that accepts a bucket folder, and converts all audio and video files in each species folder to .wav

I have placed it in our local library in the hopes that it can be made into a script and called globally later


Q-Cooke (approved):
Cool, thanks mate

PR #266, opened 2023-09-12
EE/MH/v2-Convert-2-WAV-Pipeline-notebook

mic-0-hal:
debugged
It is currently in format to convert .wav to .mp3
This is a quick patch so .wav audio files can be converted for sharing.

I will upload a more user friendly version later


Q-Cooke (approved):
Whoa this is a big file lol, cheers mate

mic-0-hal:
Should have cleared the output. It is in dire need of refactoring but it completes the task

PR #287, opened 2023-09-16
EE/AG/TL-SoundNet - Implement SoundNet Model

abhitgahlot:
In this notebook I have tried to implement SoundNet model for transfer learning with our existing model but I was not able to find it for our implementation. I have found a npy file file is used to store model parameters (weights and biases) separately from the model architecture itself and tried to implement it but it seems it has some different input and output parameters than our existing model


Ank9812 (changes requested):
I see you're using the hydrid approach here , using soundNEt with efficientNEt thats fine, 
just write short summary of this approach in the end and try to solve that ValueError you're getting in the training phase if you can

PR #292, opened 2023-09-17
EE/SWE/Add-visualisation-of-the-feature-maps

939781369:
Add a visualisation of the feature maps(VGG16)


mic-0-hal (approved):
Looks great. The next step is to incorporate our engine pipeline layers into this diagram too

PR #296, opened 2023-09-17
Untangled the components into dynamic servers.

itsRohit47:
The components built before were very tightly coupled, which means that there was never given a thought of how we would serve the echo frontend over the internet, for the end user. how will components scale, and how will debugging be done? it was all set up to be run on your PC the aim of this folder is to give a decoupled way to think of servers being entities, which are environment-dependent, what if the database server endpoint changes during the production? will stop everything and start over and go into the code and hardcode stuff? no, we use configurations that are provided wherever the component is running asking these questions is very important in real world. how about we add new features to the front end and all you have to do is commit to the branch, The rest is automation, how about releasing versions of your software? 

This folder is set up in a way, that things are entities here, and things are loosely coupled here.


Q-Cooke (approved):
LGTM, thanks heaps

PR #306, opened 2023-09-18
EE/CP/species-metadata-WIP

cpavia:
Script that reads audio files located in a specified directory and computes their durations in seconds. It then creates a metadata file in CSV format that contains information about each audio file, including its filename, duration in seconds, and class.

WIP - pushing so others can review and contribute to extend output in csv.


Q-Cooke (approved):
LGTM thanks heaps

PR #319, opened 2023-09-20
EAPI/SB/JWTHeader

CheanseanBo:
Converted functionalities to using JWTBearer instead of manually inputting token


VietTung:
Merged to hmi-dev-15, will close this request

PR #333, opened 2023-09-22
Fix x-labels in bar graph

rp097:


rp097:
Tested this version with 6 species and it works.

PR #340, opened 2023-09-22
EE/TP/Random Sampling of n samples to extract them.

ptanmay143:
[Yesterday 11:06 pm] MICHAEL HALLIDAY

Can please have a notebook written that:

accepts the input directory with all our species folders
create another output directory with identical folder structure
randomly picks 3 files from each folder in input directory
removes these and puts them in their folder in the output directory

https://teams.microsoft.com/l/message/19:a69b054085d24f289d2360ab5c59193f@thread.v2/1695387967895?context=%7B%22contextType%22%3A%22chat%22%7D


RyanBowenDeakin (approved):

RyanBowenDeakin:
@mic-0-hal this looks good to me but you might want a second look since you requested it?

PR #342, opened 2023-09-22
EE/DP/Random_Sampling

DonalPereira:
This code will allow you to randomly select the number of files needed for testing. The Input directory has to be provided from where the folders will be selected and the files within them will be randomly selected and stored in the output folder with the same folder name.


mic-0-hal (approved):
Works a treat. Thanks for making this at such short notice!

DonalPereira:
Closing this PR and making a new one due to conflicts arising because of an upload of the wrong Readme file.

PR #376, opened 2023-09-23
EE/DP/Pest_Species

DonalPereira:
1. The data/audio was scraped from Youtube using scraping code.
2. Event segmentation was performed on each file for identifying events in the audio and saved as multiple outputs.
3. Manual cleaning was performed and stored as zip folder and shared through google drive.


mic-0-hal (approved):
looks good to me

PR #382, opened 2023-09-26
the newly discovered animals i.e. the vocalisation layer is now layer…

THE0ARYANDUGGAL0experience:
…ed on top of the old animals layer and the microphones layer


Q-Cooke (changes requested):
Hi Aryan, this is a good start but there is another issue here that needs to be fixed, hmiState.layerpool needs to be initialized, I will explain further on teams

Q-Cooke:
Thanks for this Aryan

PR #396, opened 2024-02-29
Update the version for the dependencies

AndyWanng:
Updated the version of the dependencies below:
- paho-mqtt
- Werkzeug


dipanshgarg21 (approved):
Version change of paho-mqtt to 1.6.1 and added Werkzeug version to 2.3.7

PR #435, opened 2024-04-14
Convert_h5_to_pb

Joel213141:
notebook to convert tensorflow .h5 models to keras .pb as well as a code to test if the .h5 file is appropriately converted to .pb format.


ptanmay143 (commented):
Hi @Joel213141. It seems that you have accidentally pushed the .ipynb_checkpoints instead of the actual integration notebook with outputs.

PR #440, opened 2024-04-17
Unsupervised bird modelling

jacob7700:
Performed unsupervised learning using a specific bird dataset and implemented prediction model based on clustering.


AndyWanng:
Morning @jacob7700 , I've browsed all your script yesterday and it's good to see that you are willing to explore different methods on unsupervised learning and ensemble learning, but I'm afraid your code is not well organized and it's not good enough to be merged into the main. Here're several aspects you can work on:
1. Since we already have an individual script to convert audio files, I think you can exclude this method from your script to avoid redundancy.
2. It's good to see you want to do some work on Exploratory Data Analysis(EDA), however, instead of listing all of the features extracted from audio, it would be better if you could add the method to analyse those features.
3. For unsupervised learning, K-means is a very normal approach to apply this, but you only used sample rate and duration to for clustering which doesn't make much sense as you have set the sr to 16000 in your previous method and it won't be of much help if you cluster those data only by the differences between the duration time.
4. Good to see you used AdaBoost from scikit-learn, but you only built this as a very simple structure which only used mfcc feature for classifier which I doubt whether it would be capable of audio classification tasks considering animals would have different patterns, frequencies, etc. for their sounds.
5. You used jupyter notebook for your script but it's not well organized and there are a lot of redundant functions. It would be much better if you could reorganize your code to make it more readable and help it maintain a clean structure.
It's good to see you trying to delve deeper for those advanced tasks, however, since each task is a very complicated one, I would suggest you focus on one of them ATM, for example, EDA task, so that you can enhance your code with higher quality and complexity instead of using a very simple structure to get very limited results from the audio files. Please feel free to reach out to me if you have further questions.

AndyWanng:
Also, could you please remove one of your files named "d" from your PR in your next commit? I believe you must've uploaded it to the repo by accident.

PR #582, opened 2024-09-19
Added weather pipeline function to echo engine.py

kjkombat:
Adds only weather pipeline function to engine.py for integration of weather model

This function is not being used and doesn't effect the current main loop.


empaid (approved):
Discussed with KJ on meet

PR #616, opened 2024-09-27
Update file location

minhD03:
I have updated file location for my sprint 2 task.


regina-arissa (changes requested):
this is an old commit, but good quality report, please move the folder to the designated place (in src prototype, and name the folder to the task name)

PR #630, opened 2024-11-22
Backlog Sprint1- Region-Based Mapping

mariaelsa14:
Map animals to their respective regions based on uploaded images, using a dictionary for region mapping. The system will process uploaded media, predict the animal, and associate it with a predefined region.


regina-arissa:
Why merge with your onboarding branch instead of merging to main?

PR #640, opened 2024-12-07
Sprint 2 Task: Course Detection

minhD03:


src/Prototypes/engine/Course Detection - ND/Course Detection - ND.ipynb
regina-arissa, line 1: Although you did a great job in doing these tasks and providing the report, it would have been better if you had provided more explanation on each functions that you implement. Because it took me a while to review your codes and understand what each of the function does.

src/Prototypes/engine/Course Detection - ND/Course Detection for Node device -  ND.pdf
regina-arissa, line 1: The report is well written, and you have provided great details in the document. However, would be better if you had provided a short summary in the beginning of the document of:
1. what data you use
2. the method used to identify animal sounds
3. result

Also, even though you use our original data to detect whether the sound is coming from an animal or not, I think you did not use any environmental sounds as the opposite. Thus might be better for your future analysis if you could include animal sounds (as your positive control) and environmental sounds (as your negative control). This makes it easier to determine how well your model works in identifying whether the sound comes from an animal or not

regina-arissa (approved):
Overall you did great, just need a few improvement for your next reports:
1. provide function explanation
2. provide more test samples that you know is not an animal sound and see if the model works well too

PR #645, opened 2024-12-11
Ee/ib/movingsprint1and2totherightdirectory

IrfanBoen:
I had moved both of my Sprint 1 and Sprint 2 task to the right directory which is at C:\Users\User\Project-Echo\src\Prototypes\engine


regina-arissa:
are you sure you have moved it instead of just copy and paste? because I see no deletion, only addition for this changes. And the addition is for both in the correct folder, and the incorrect folder

IrfanBoen:
Yes you are right, sorry about that. I will try and pull a request again

IrfanBoen:
I had just pull a request that has fixed the problem. I had made sure that the files in the wrong directory had been deleted and made sure that they exist in the right directory only.

PR #655, opened 2024-12-12
Tasks after changes

mariaelsa14:


mariaelsa14:
I've made some changes to make the region based mapping look more understandable.

ML217:
Long overdue task

PR #664, opened 2025-01-08
Add files via upload

RishWork:


mahajokhio2 (changes requested):
Hi Rishabh, 

I went through your code, do you mind just writing up a short comparison report on how this is faster in terms of runtime to the other pre-existing optimised model?

I saw that you also used mixed precision and it's meant to print when mixed precision is enabled but I don't see the print statement in the output. I also noticed that the code with the mirrored strategy and mixed precision is repeated twice. It would be good to discuss the changes either in the report or in a markdown cell within the notebook.
 
Lastly, I noticed that you used profiler as well which is a good strategy for recognising bottlenecks in the model, maybe try utilising the profiler and talking about the bottlenecks you identified in your report. It might be helpful for future students.

ML217:
overdue

PR #665, opened 2025-01-09
Runtime Optimisation

mahajokhio2:


bevanyeah (approved):
All changes related to optimization strategies in notebook form, which are all useful to keep in the main branch going forward.  Thanks! approved.

PR #670, opened 2025-03-18
Updated Onboarding Task

brayden276:


mahajokhio2 (changes requested):
Hey Brayden, there's a few steps missing from your onboarding task such as the screenshot that needs to be inserted. I also noticed that the rendered cells have folder names from my computer which shouldn't be the case since you are doing the task on your computer with different names for the folders, I presume. If you could re-upload the completed task, that'll be great!

ML217:
Cannot view committed changes.

PR #719, opened 2025-04-30
Updated model pipeline for frequency augmentation and testing

Jessica-Stinson:
The model pipeline has been updated to include frequency augmentation during training and enhance the repeatability of model testing. The splits used during training and testing of the model have been included and can be loaded into the notebook for further testing.


mahajokhio2 (approved):
This is really good work Jess! Thank you!

PR #721, opened 2025-05-06
Add live event log with map highlight integration

rsanada29:
### Summary
- Added live events log UI to dashboard
- Clicking on a log entry highlights the microphone location on the map
- Connected to backend `/hmi/events_time` API
- Used Leaflet.js for map rendering
- Added new data files under `public/io`


ML217:
has conflict

PR #728, opened 2025-05-15
Enquiry feature

kzthwe:
Implemented enquiry model, connected with the backend API, and added Redis support


ML217:
has conflict

PR #739, opened 2025-05-20
Engine Model Optimisation 

RishWork:
Engine Model has been further optimised with documentation for future enhancements. 
[Optimised Model Report.pdf](https://github.com/user-attachments/files/20344805/Optimised.Model.Report.pdf)

### DIFF
--- a/src/Prototypes/engine/Engine_Model_Optimization.ipynb
+++ b/src/Prototypes/engine/Engine_Model_Optimization.ipynb
@@ -1,0 +1,698 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


mahajokhio2 (changes requested):
Hi Rishabh, looking through your work, it seems that you are trying to optimise a model on simulated melspectograms when the actual task itself called for optimisation of the tensorflow pipeline already available in the repository that actually utilises our dataset. Please review and update, if by any chance, I have misinterpreted your woke, let me know! I can't merge this with the repository at its current state.

PR #753, opened 2025-07-27
[fix]: for why other students cannot login the Echo Localhost

Tennie002:
In the main.py someone commited  unresolve  Merge Conflict.

- It was missing the packages to run the code inside the main file. I added the auth_router and Json packages
- In Auth_router, I changed the module with correct path because the docker just given every under app folder.


DeanFawcett (approved):
Good work in finding this error. I have run a test and the docker container compiles.

PR #769, opened 2025-08-29
Torch training pipeline

newbee1905:
Implement training pipeline in torch 

Supporting
- Configurating augmentation through yaml file
- Configurating models
- Builtin custom models with arcface support
- Support QAT and Post Quantisation
- Still require more testing for Knowledge Distilation (will finish this within sprint 2)


ML217 (approved):
looks fine

PR #878, opened 2026-03-31
Connect sensor dashboard to live MQTT payload

ohwenyi:


mraula (approved):
everything looks good

PR #880, opened 2026-04-02
Ee/kfw/benchmarking task1

KishanFW:
benchmarking done for four models. all information there in the file. After this, Jack did some changes to his task.


mraula (approved):
Everything looks good.

PR #888, opened 2026-04-13
Add IoT MQTT integration for engine with unit tests

mraula:
Adds echo_engine_iot.py with fixed IoT handler from IoT team review:
- Payload validation (rejects missing audio/GPS)
- Reuses echo_api_send_detection_event instead of duplicated logic
- Reads sensor_id and gps_uncertainty from payload
- IoT listener starts after class_names loaded (race condition fix)
- Broker/topic/port configurable via echo_engine.json

Includes test_iot_integration.py (14 unit tests) and test_iot_publisher.py (manual MQTT test script).


KishanFW (approved):
Eveyrthing is good.

PR #889, opened 2026-04-14
Added Benchmarking Report for T1 2026

jackhatesgravity:
Using the benchmarking environment, I've run through all the required models, calculated the necessary performance metrics, and written up a report detailing the benefits of some of our selected model architectures over others.


mraula (approved):
everything looks good

PR #893, opened 2026-04-23
Task5(API cofigurable): Update auth.routes.js

kimmynoto:
Task 5 (API configurable): I have updated the line 224 setting it as a live server when the user forgot the pw so it's ensure the reset password work locally and on live server as well


mraula (approved):
everything looks good

PR #896, opened 2026-05-01
Ee/wkf/efficientnetv2 integration

KishanFW:
I integrated the selected EfficientNetV2 TFLite model into the Engine-side inference path and validated preprocessing, class mapping, confidence output, and prediction.
The model now loads successfully inside the Docker Engine container with input shape [1, 1, 128, 313] and output shape [1, 123]. I also tested a real audio file inside the container, and it correctly predicted Acanthiza chrysorrhoa with 93.37% confidence.
Current blocker: the Engine reaches the MQTT connection stage but cannot resolve mqtt-broker:1883.
Error: CONNECTION ERROR: [Errno -2] Name or service not known
So the model side is working; the remaining issue seems to be Docker Compose/MQTT service-name or network configuration.


mraula (approved):
everything looks good

PR #898, opened 2026-05-04
EE/RM: Fix MQTT broker hostname and missing deps in light engine

mraula:
## Summary

- **light_echo_engine.json**: Corrects MQTT_CLIENT_URL from "mqtt-broker" to "ts-mqtt-server-cont" which is the actual Docker Compose container name. The wrong hostname caused [Errno -2] Name or service not known on every connection attempt, preventing the light engine from starting.
- **
Requirements.txt**: Adds four packages imported by light_echo_engine_efficientnetv2_tflite.py that were missing, causing the container to crash on startup:
  - paho-mqtt==1.6.1
  - pymongo
  - geopy
  - google-cloud-storage


KishanFW (approved):
Eveyrthing is good

PR #899, opened 2026-05-06
Validate EfficientNetV2 MQTT inference in Docker

KishanFW:
I updated the Engine inference flow to use the local EfficientNetV2 TFLite prediction path instead of the old TensorFlow Serving `outputs` format. The Engine now produces predicted species labels, confidence values, and top predictions from MQTT audio messages.

## Files Changed

- `light_echo_engine_efficientnetv2_tflite.py`  
  Updated MQTT audio inference to use EfficientNetV2 TFLite locally.

- `system_manager.py`  
  Fixed simulator start mode handling.

- `control_sim.py`  
  Updated simulator testing flow for `Animal_Mode`.

- `comms_manager.py`  
  Added debug checks to confirm valid WAV audio was being sent.

- Engine requirements file  
  Updated Engine dependencies needed for Docker/TFLite inference.

- Simulator requirements file  
  Updated Simulator dependencies needed for MQTT/audio testing.

- `README_EfficientNetV2_Engine_Integration.md`  
  Updated final Docker/MQTT validation status.

## Note
The remaining `Internal Server Error` happens after prediction when sending results to the API/database, so it appears to be an API/backend issue.


mraula (approved):
everything looks good

PR #900, opened 2026-05-08
task-2-1-detect-multimodal-failure

jayden-s222587117:
Changes:
- Added MultiModalPredictionError
- Added prediction response validation
- Moved placeholder prediction logic into model_adapter
- Updated /predict to return a controlled 502 error when model output is invalid


mraula (approved):
everything looks good

PR #904, opened 2026-05-14
Added error_analysis

jackhatesgravity:
Added error_analysis folder with analysis notebook and resulting output that covers core essentials of production model strengths and weaknesses.


KishanFW (approved):
Everything is good

PR #906, opened 2026-05-16
feat(frontend): Task 7 - shared UI utilities, loading states, retry handling

Vamshi-Gollapelly:
## What this does

This update continues the frontend reliability work from Sprint 2 
and extends it to more parts of the app.

### HMI-utils.js
Added several reusable helper functions that other parts of the 
frontend can now use:

- showToast() now supports four types: success, error, warning, 
  and info. Each one has a different colour so users can quickly 
  tell what kind of message it is. Toasts also have a close button 
  and auto-dismiss after a few seconds.

- withRetry() wraps any API call and automatically tries again if 
  it fails. It shows a warning toast between each attempt so the 
  user knows what is happening.

- getApiErrorMessage() takes an error from a failed request and 
  returns a short, plain-English message instead of showing a 
  raw error or stack trace to the user.

- showPageBanner() and hidePageBanner() show a coloured bar at the 
  top of the page for things that need to stay visible, like a 
  critical alert. Unlike toasts, these do not disappear on their own.

- confirmDialog() is a simple yes/no popup that works like 
  window.confirm() but does not block the browser. It returns 
  true or false depending on what the user clicks.

- guardRequest() stops the same request from being sent twice if 
  the user clicks a button more than once before the first one 
  finishes.

- debounce() stops a function from firing on every single keystroke, 
  which is useful for search inputs.

- guardFormSubmit() disables the submit button while a form is 
  being sent, shows a loading label, and then shows a success or 
  error toast when done.

### routes.js
- getApiErrorMessage was defined in two places with the same code. 
  Removed the copy from routes.js so there is only one version.
- All GET requests now use withRetry so they automatically try 
  again on a network failure before showing an error to the user.

### map.js
- Removed several local functions that were doing the same job as 
  the shared helpers in HMI-utils.js. The file now just imports 
  and uses the shared ones instead.

### index.html
- The app had its own custom notification div and a generateVisualResponse() 
  function that did the same thing as showToast(). Removed the 
  custom one and replaced all of those calls with showToast().
- Also replaced all alert() popups with toasts so the experience 
  is consistent across the whole app.
- Fixed how HMI-utils.js was being loaded so its functions are 
  actually available to the rest of the page.

### middleware/index.js
- The session check had its logic the wrong way around. Routes 
  like /map that should be public were requiring a login, and some 
  protected routes were being skipped entirely. Fixed this so 
  public and protected routes are clearly separated.
- Added a clearUserSession() function for use in logout flows.

## Bug fixes
- map.routes.js was calling next() after already sending a response, 
  which caused the server to crash. Removed those calls.
- The same crash was happening in user.routes.js. Fixed the same way.
- server.js was crashing when the Python API was unreachable. 
  Added a check so it handles that gracefully instead.

## What it looks like
- Toasts appear bottom-right with the correct colour for each type
- Retry attempts show as warning toasts counting up (attempt 1/3, 2/3)
- The server no longer crashes when a request fails
- All API calls proxy correctly through the Node server to the Python backend


KishanFW (approved):
Everything is good

PR #917, opened 2026-05-17
Standardized how prediction API should be responed

ML217:


KishanFW:
For repository restructure

PR #920, opened 2026-05-19
HMI - Spectrogram Implementation 

gorillaPet:
Implemented spectrogram and default audio pathing for live sensor data to display spectrogram.


mraula (approved):
everything looks good.

PR #925, opened 2026-05-31
Fix(Ee)/rm/iot engine integration

mraula:
fixed integration errors after unit testing

- test_iot_integration.py: fix patch target from echo_engine to echo_engine_iot
- test_iot_integration.py: update MODEL_SERVER and WEATHER_SERVER URLs to match
  echo_engine.json, add missing API_URL field
- All 16 unit tests now pass


KishanFW (approved):
Everything is good

PR #926, opened 2026-07-19
Krish fix issues T2 2026. 

KishanFW:
I fololowed the debug guide shared by jamie and some my inpouts to fix initial errors poped up when setting up echonets.


seudhipa404 (approved):
Reviewed the dependency and EchoNet setup fixes. The changes look good to me.

PR #927, opened 2026-07-19
Fix Echo engine connection

zeinware:
Updated the docker-compose.yml

on the echo_engine services
updated context to ./Engine
updated dockerfile to Engine.Dockerfile


KishanFW (approved):
all good

PR #935, opened 2026-07-28
Restructure repository into team-based layout (Krish's structure)

Mangon3:
## What this PR does

Reorganizes the repository into the agreed team-based structure ("Krish's structure"),
separating production code, prototypes, shared tooling, deployment, and tests, and
moving docs/tutorials/data into clear top-level homes. All moves were done with
`git mv` so file history is preserved.

This is a full-repo relayout, so the diff is large by nature (~16.6k files). Please
review the **structure and the mapping below**, not the line-by-line diff.

## New layout

```
src/
  production/       # deployed services: backend, engine, hmi, iot, simulator, infrastructure
  prototypes/       # experimental work: backend, engine, hmi, iot, simulator, computer_vision
  data_tools/       # shared data scripts, training, utils
  deployment/       # docker/, kubernetes/
  tests/            # load / integration tests
docs/               # architecture/, team-guides/, research/
tutorials/          # onboarding task + guides
models-and-data/    # datasets / models (DVC-tracked)
```

## Key changes

- `src/Components/*` production services -> `src/production/*` (lowercased, per-team).
- Experimental notebooks/tasks -> `src/prototypes/*` with themed engine subfolders
  (pytorch, augmentation, event_detection, benchmarking, weather_noise, ensemble_overlap).
- Docker/K8s configs consolidated under `src/deployment/`; compose build contexts
  updated and validated with `docker compose config`.
- Python package relocated; `setup.py` `package_dir`/`find_packages` updated.
- `.github/CODEOWNERS` remapped to the new paths (broad tree defaults ordered before
  specific rules so last-match-wins resolves correctly).
- Migration mapping documented in `docs/architecture/MIGRATION_KRISH_STRUCTURE.md`.

## About the old archive (`.delete/`) - please read

Heads up: this PR **intentionally includes** the `.delete/archive/` tree - roughly
**3.6 GB across 8,619 files** (the former `src/Archive`: old `src/Components`,
`Prototypes/`, `Echo_Components_on_K8s/`, etc., including some 80+ MB model blobs).

It is kept **on purpose and temporarily** so the whole team can relocate their own
work out of it *in the shared repo* rather than blind-deleting anything.
`docs/team-guides/DELETE_ARCHIVE_MIGRATION_GUIDE.md` tells each member how: move files
into existing folders only (no new folders, no overwrites, `git mv` to preserve
history), and delete regenerable junk instead of moving it.

**Consequence:** merging this enlarges the repo/clone by ~3.6 GB until the migration
[...]

PR #937, opened 2026-07-29
fix(docker): restore missing echo_engine service to docker-compose.yml

Mangon3:
## Problem

After the restructure PR (#935) merged, team members reported the **engine container was missing** — `docker compose up` no longer creates `ts-echo-engine-cont`.

## Cause

The `echo_engine` service was accidentally deleted from `src/deployment/docker/docker-compose.yml` during the restructure, in commit `010cfdf3` ("restructure repository in chunks"). It went from present to gone there and stayed gone through the move + merge.

It survived only in `docker-compose.test.yml` (which is why CI still counted it and the loss went unnoticed until someone ran the **normal** stack). `model_server` (TF Serving, `ts-echo-model-cont`) is still present, so it *looks* like the engine is there when it isn't.

## Fix

Restore the `echo_engine` service, one service block, retargeted to the restructured engine dir:

- `context: ../../production/engine`, `dockerfile: Engine.Dockerfile` — matching `model_server` and the test compose (the original pointed at `Prototypes/engine/torch_impl` / `light_engine.Dockerfile`, now deprecated and moved out).

Validated with `docker compose config`: the stack is back to 10 services and `echo_engine` resolves correctly.

## Note (separate, not addressed here)

Actually *running* the engine still needs GCloud credentials — its entrypoint runs `gcloud auth application-default login` then `python echo_engine.py`. This PR only restores the missing service so the container exists in the normal stack again; the credentials/boot behaviour is a separate follow-up.

### DIFF
--- a/src/deployment/docker/docker-compose.yml
+++ b/src/deployment/docker/docker-compose.yml
@@ -20,6 +20,19 @@ services:
     stdin_open: false
     tty: true
 
+  echo_engine:
+    build:
+      context: ../../production/engine
+      dockerfile: Engine.Dockerfile
+    image: ts-echo-engine
+    container_name: ts-echo-engine-cont
+    networks:
+      - echo-net
+    volumes:
+      - credentials_volume:/root/.config/gcloud/
+    stdin_open: true
+    tty: true
+
   echo_hmi:
     build:
       context: ../../production/hmi


KishanFW (approved):
Everything is good

PR #943, opened 2026-08-01
B3.2: Fix MongoDB persistence for engine event confidence

HarryTrieu:
## Summary
- Change EventSchema.confidence from condecimal to confloat so PyMongo can persist engine events.
- Targets current flow: POST /engine/event -> EchoNet.events.

## Evidence
Sprint 1 evidence portfolio:
https://docs.google.com/document/d/1PgN9Bc4MURyD_nz11zhZwj_YqldjuhUc-X6f_wPiUuE/edit?usp=sharing

## Test plan
- [ ] POST /engine/event with confidence 95.0 returns HTTP 201
- [ ] Document appears in EchoNet.events with matching _id
- [ ] Same document survives MongoDB container restart
- [ ] confidence 0 and 100 still return HTTP 422
- [ ] Delete only the temporary test document after verification


HarryTrieu:
Conflict resolved by rebasing onto latest `main`.

`main` already contains an equivalent persistence fix:
`confidence: float = Field(gt=0, lt=100)`

That replaces the old `condecimal` path the same way my original `confloat` change did, so this PR no longer has unique commits. Safe to close as superseded by the current `main` schema, unless reviewers want this PR kept for history.

PR #944, opened 2026-08-01
C2.3: Add MongoDB mocking for local API tests

HarryTrieu:
﻿## Summary
- Add requirements-test.txt, mongomock fixture, and POST /engine/event integration test.
- Patch MongoClient at conftest import time so mocking works for the full `pytest tests/` suite (including module-level app imports such as `test_schema_validation.py`).
- Depends on B3.2 confloat fix for BSON-compatible confidence values.

## Evidence
Sprint 1 evidence portfolio:
https://docs.google.com/document/d/1PgN9Bc4MURyD_nz11zhZwj_YqldjuhUc-X6f_wPiUuE/edit?usp=sharing

## Test plan
- [x] Install requirements-test.txt in the API container
- [x] `pytest tests/test_mongo_mock.py` passes
- [x] `pytest tests/` (whole folder) passes in the Docker API container (36 passed)
- [x] Mock takes effect even when other tests import `app` at module level during collection
- [x] Same mock path does not require a live MongoDB connection for the C2.3 test


HarryTrieu:
Rebased onto latest `main`. Conflict in `schemas.py` resolved by keeping `main`'s `float = Field(gt=0, lt=100)`. This PR now contains only the C2.3 test files.

KCP17 (changes requested):
I checked this out locally and ran it alongside `test_schema_validation.py` (also in `tests/`). That file loads the app at module level, and pytest imports every test file during collection before any fixture runs, so the real database connection gets created before your mocking fixture ever runs. When your fixture tries to patch it afterward, Python just returns the already-loaded module instead of re-running it, so the mock never actually takes effect. Your own `assert isinstance(...)` check will catch this and fail loudly, which is good, but it means this test currently only passes in isolation, not as part of the full suite. Can you run `pytest tests/` (the whole folder, not just this one file) inside the Docker container to confirm? If it fails like I expect, the fix is to make the mocking patch happen earlier, before any test file gets loaded at all (a fixture that runs first automatically for the whole session), or better, fix it upstream so the database module doesn't connect to MongoDB immediately just from being imported.

HarryTrieu:
﻿Thanks @KCP17 — confirmed and fixed.

**Root cause**
You were right: `test_schema_validation.py` (and other files) import `app` at module level during pytest collection, which is before any fixture runs. `app.database` opens a real `MongoClient` on import, so the old fixture monkeypatch never replaced the already-bound collections.

**Fix (92895086)**
- Patch `pymongo.MongoClient = mongomock.MongoClient` at **conftest import time** (plus `pytest_configure` safety net), before test modules are collected.
- Set required Settings / mail env placeholders in the same early conftest path so the full suite can import `app.main`.
- Updated C2.3 / schema payloads for the current `sourceType` + LLA object contract.
- Lazy-init `DetectionStreamManager` lock so Python 3.9 does not blow up when the manager is constructed outside a running loop under pytest.

**Verification (inside Docker API container)**
```text
pip install -r requirements-test.txt
pytest tests/ -q
# 36 passed
```

That run includes `test_schema_validation.py` and `test_mongo_mock.py` together, and the mongomock `isinstance(...)` assert still passes.

Ready for re-review.

KCP17 (approved):
I pulled this branch into a clean environment and ran pytest tests/ -q myself, 36 passed, matching your count. The conftest-import-time patch is the right fix, it runs before pytest collects any test module, which is exactly the window the old fixture-based monkeypatch missed. test_schema_validation.py and test_mongo_mock.py both pass together, and the mock_mongo_api fixture's own isinstance check on the Mongo client held too. I'm approving this, good work!

PR #946, opened 2026-08-01
FR-D1: Add shared API client with consistent error handling

Alexander-Petkovski:
## Summary
Adds one shared API client (`services/apiClient.js`) for all HMI Node-side calls to the
Python API, replacing three separate copies of `axios` + `try/catch` that each handled
errors differently.

## Why
Before this change:
- `API_BASE_URL` was computed independently in three files (`server.js`,
  `auth.routes.js`, `map.routes.js`).
- Every route rolled its own error handling, with inconsistent results - some
  returned `502`, some `500`, some `404` for the same kind of failure.
- The signup route in `auth.routes.js` would actually crash with a `TypeError`
  if the API was simply unreachable, because it assumed `err.response` always
  existed on a failed axios call.

`apiClient.js` fixes this in one place: it always tells apart a backend error
response, an unreachable backend (network/timeout), and a malformed request, so
every route gets the same reliable behaviour instead of a different guess each time.

## What changed
- **New:** `src/production/hmi/ui/services/apiClient.js`
- **Migrated to use it:**
  - `routes/map.routes.js` - 8 routes (movement/events/microphones/audio/recording/sim control)
  - `server.js` - 5 routes (`/api/submit`, `/api/requests` GET+PATCH,
    `/api/updateConservationStatus/:animal`, `/iot/nodes`)
  - `routes/auth.routes.js` - 5 routes (signup, signin, 2fa verify, forgot
    password, reset password)
- **Deliberately left alone:** `proxyToApi` in `server.js` - it needs to forward
  whatever status the backend returns (including error ones) via
  `validateStatus: () => true`, which is the opposite of the shared client's
  throw-on-error behaviour. It does now reuse the same `API_BASE_URL` though.
- Removed a few `else` branches that checked for an exact success status code
  (e.g. `=== 201`) - these were dead code, since axios already throws on any
  non-2xx response, so the branch could never actually run.
- Kept one specific behaviour intentionally: the signin route still special-cases
  a `401` response to show "Invalid credentials" rather than a generic error,
  since that's a real, reachable case (confirmed against `auth_router.py`).

## Testing
Tested against the real local `docker-compose` stack, not just a syntax check:
- **Success path:** `GET /microphones` returned real MongoDB data (`200`).
- **Failure path:** stopped the `ts-api-cont` container mid-session, hit the same
  route - got a clean `502 {"error":"API unavailable"}` instead of a hang/crash.
  Restarted the container, confirmed it recovered to `200`.
- **Auth error path:** signin with a non-existent user surfaced a clean
  `404`-based message with no crash.

## Open question / not in scope here
This covers the **server-side** Node layer only. The **browser-side** `fetch()`
calls (`public/js/*.js`, `user-visits.html`, etc.) still have no shared client -
unclear yet if that's part of FR-D1 or a separate follow-up. Flagging for
reviewer input.
^ TLDR is it my job?


src/production/hmi/ui/routes/auth.routes.js
```diff
@@ -58,66 +52,57 @@ module.exports = function (app) {
...
-      console.log('Login exception error: ' + err)
-      res.send(`<script> window.location.href = "/login"; alert("Login exception Error: ${err}!");</script>`);
-    }  
+      console.log('Login exception error: ' + err.message)
+      // 401 from the API means bad credentials specifically, not some other failure - worth telling the user that outright
+      const alertMsg = err.status === 401
+        ? 'Failed! Invalid credentials!'
+        : `Login exception Error: ${err.message}!`;
+      res.send(`<script> window.location.href = "/login"; alert("${alertMsg}");</script>`);
+    }
```
AnDo081105, line 118: Problem: this catch returns HTTP 200 HTML, but `login.html` treats any 2xx response as JSON success. It then fails during `response.json()`, so the intended invalid-credentials message is never shown. Return a non-2xx JSON response matching the browser contract, and avoid exposing internal error details.

```suggestion
    } catch (err) {
      console.error('Sign-in failed:', err.message);

      if (err.status === 401 || err.status === 404) {
        return res.status(401).json({
          message: 'Invalid username, email, or password.',
        });
      }

      if (err.isNetworkError) {
        return res.status(502).json({
          message: 'The sign-in service is currently unavailable.',
        });
      }

      return res.status(500).json({
        message: 'Unable to sign in. Please try again.',
      });
    }
```

AnDo081105 (changes requested):
Requesting changes for the remaining FR-D1 gaps:

1. The implementation centralizes only Node-side Axios calls. FR-D1 requires the shared access pattern across HMI and admin pages, including consistent read-only retries and standard user-facing error states. Page-local `fetch`/Axios calls remain in login, admin nodes, admin requests, sensor health, API status, and node overlay code. Please reuse or extend `public/js/routes.js`, migrate those calls, retry only safe read-only requests, and route failures through the shared banner/toast utilities. If this PR is intentionally Node-side-only, it should be re-scoped and linked to a follow-up rather than marked as completing FR-D1.

2. This shared request/error layer has no automated regression coverage, and the package `npm test` script currently exits with "Error: no test specified". Please add focused tests for response parsing, backend 4xx/5xx normalization, network/timeout handling, safe user-facing messages, GET retry behavior, and confirmation that state-changing requests are not retried.

The core client passed focused success/backend-error/network-error checks, and all four changed files passed `node --check`.

Vamshi-Gollapelly (changes requested):
I went through the changes and don't think this PR is ready to be approved yet. The sign-in error handling still returns HTTP 200 with HTML, while the browser expects a JSON response, so the error isn't handled as intended. There are also a few FR-D1 requirements that are still missing, including the browser-side shared API/error handling and the automated test coverage. Please address these before this PR is approved.

Alexander-Petkovski:
Thanks — all three points are addressed in 7c9506d.

## 1. Browser-side shared client

All six areas you named now go through `public/js/routes.js`:

| File | Was | Now |
|---|---|---|
| `public/login.html` | raw `fetch('/api/auth/signin')` | `signIn()` |
| `public/admin/admin-nodes.html` | `fetch('http://localhost:9000/iot/nodes')` ×2 | `retrieveIotNodes()` / `retrieveIotNode()` |
| `public/admin/admin-request.html` | 4 raw `fetch` calls | `retrieveAdminRequests()`, `updateRequestStatus()`, `updateConservationStatus()` |
| `public/admin/sensor_health/script.js` | its own `apiFetch()` helper, 6 call sites | 6 new `routes.js` functions |
| `public/js/api-status.js` | raw `fetch('/iot/nodes')` | `retrieveIotNodes({ silent: true })` |
| `public/js/nodes-overlay.js` | `axios.get('/iot/nodes')` | `retrieveIotNodes()` |

Also picked up `sensor_health/charts.js`, which was the last raw `fetch` left on the sensor health screens.

Retry policy follows the existing convention in `routes.js` GETs wrapped in `withRetry` (3 attempts), POST/PATCH/PUT never retried, each with a comment on why. Failures route through `getApiErrorMessage` into the existing toast/banner utilities.

Two parts of the diff worth explaining up front:

- **`axios` script tags added to eight admin pages.** `routes.js` reads `window.axios` at module-evaluation time, and none of the admin pages loaded it. Without the tag the module throws while evaluating and the importing script never runs at all.
- **New `/iot/nodes/:nodeId` route in `server.js`.** The API already implements this (`app/routers/iot.py`), but there was no Node route for it, so `admin-nodes.html` was hitting `localhost:9000` straight from the browser, which only ever worked in local dev. Same shape as the `/iot/nodes` route directly above it.

## 2. Test coverage

`npm test` now runs jest instead of the placeholder. 33 tests across three files, mapped to your list:

| Your ask | Where |
|---|---|
| Response parsing | `services/__tests__/apiClient.test.js` |
| Backend 4xx/5xx normalisation | same |
| Network/timeout handling | same |
| Safe user-facing messages | same, plus `public/js/__tests__/HMI-utils.test.js` |
| GET retry behaviour | `public/js/__tests__/routes.test.js` |
| State-changing requests not retried | same — `postRecording`, `setSimModeAnimal`, `updateRequestStatus`, `updateConservationStatus` and `signIn` are each asserted to fire exactly once |

note: `node_modules/` is partially tracked in this repo but jest isn't among it, so you'll need `npm install` in `src/production/hmi/ui` before `npm test`.

## 3. Sign-in bug

Applied your suggested shape exactly. Verified against the live docker-compose stack — bad credentials now return `401` with `Content-Type: application/json` and `{"message":"Invalid username, email, or password."}`, and `login.html` surfaces it through the shared toast. No `err.message` leaking to the client any more.

## Deliberately not in this PR

**MFA/OTP.** The `mfa_phone_enabled` branch in `auth.routes.js` has the same HTML-with-200 problem, and `login.html` has no handling for the OTP redirect. Since that's a 2FA login flow it sits with **FR-D4 (reCAPTCHA and Security Hardening)** rather than FR-D1, so I've left it out rather than widen this PR , flagging it here so it's tracked.

**One for whoever owns the API:** `app/routers/iot.py` raises its own `HTTPException(404, "Node not found")` inside a `try` with a bare `except Exception`, so an unknown node id comes back as a `500` reading `"Error retrieving node: 404: Node not found"`. Same behaviour hitting `:9000` directly, so it's not from the proxy.

Vamshi-Gollapelly (approved):
Pulled the branch and actually ran it instead of taking the summary at face value. Tests check out, 33/33 passing, verified myself. Sign-in fix is solid too: proper JSON errors now (401/502/500), and login.html was updated to match, so both sides actually agree. Shared-client migration is real across api-status.js, nodes-overlay.js, login.html, sensor-health checked each one.

One flag, not a blocker: nodes-overlay.js here is built on the pre-#960 version, so it's missing your dedup fix. Since both PRs are open at once, whoever merges second will need to carry both fixes through the conflict — worth a quick comment so it doesn't get lost.

AnDo081105:
Please finish the FR-D1 browser migration by adding shared-client methods for notifications, submissions, profiles, and donations, then replace the remaining production-facing raw API calls. Keep static asset loading and the API explorer as documented exceptions. Relative browser routes should replace direct localhost:9000 calls, GET retries should be limited to transient failures, and state-changing requests must remain single-attempt.

Alexander-Petkovski:
Round two is done in d3365dd.

## Shared-client methods for the four areas

| Area | Methods added | Call sites migrated |
|---|---|---|
| Notifications | `retrieveUserNotifications`, `addUserNotification`, `removeUserNotification` | `index.html` ×3 |
| Submissions | `submitAnimalRequest` | `index.html` |
| Profiles | `retrieveUserProfile` | `index.html` |
| Donations | `retrieveCumulativeDonations`, `retrieveDonations` | `admin/dashboard.html`, `admin/donations.html` |

## Relative routes instead of localhost:9000

All gone there are now **zero** direct `localhost:9000` calls anywhere in `public/`. The remaining string matches are comments explaining what changed.

- the three notification calls in `index.html`
- `fetchWeatherData` in `js/HMI.js`
- the three node calls in `admin/admin-nodes-temp.html`

`server.js` already proxies `/hmi/*`, so these only needed to stop naming the host. I also added `/iot/nodes/:nodeId` to `server.js` in the previous commit — the API implements it (`app/routers/iot.py`) but there was no Node route, which is why that page was going direct.

## GET retries limited to transient failures

`withRetry` now takes a `shouldRetry` predicate, defaulting to a new `isTransientError`:

- **retries:** timeouts, requests that never got a reply, 408, 429, 5xx
- **does not retry:** 4xx a 401 or 404 is a settled answer, so three attempts only delayed the error the user needed to see

Measured on the live stack: a 403 on the notifications route now fails in **8ms on a single attempt**, where it previously took three attempts over roughly three seconds and threw two retry toasts on the way.

## State-changing requests

Still single-attempt, and the new ones are covered: the notification POST and DELETE, and the submission POST. There are assertions for each in `routes.test.js`.

## Documented exceptions

Staying on raw `fetch` deliberately:

- **API explorer** (`admin/js/api-explorer.js`) — its whole purpose is issuing arbitrary requests, so a fixed-route client doesn't apply.
- **Static assets** — `js/spectrogram.js` loads an audio file and `index.html` loads `sample_data.json`. Not API calls.
- **Payments** (`index.html` Stripe checkout, `paymentgateway.js`) — these talk to Stripe/Razorpay flows rather than the Echo API, and I'm not putting a payment path through a client change in this PR.
- **`admin/feedback.html`** — single call, belongs with the feedback work rather than the transport layer.
- **`admin/test_1.html` / `test_2.html`** — scratch pages, not production-facing.

## Tests

48 jest tests plus the 5 `node:test` ones already on main. flagged: main moved `npm test` to `node --test` while this branch was adding jest, so `npm test` now runs both runners and neither side's tests get dropped. The globs also cover `tests/**`, so #960's file will be picked up automatically once that lands.

You'll need `npm install` in `src/production/hmi/ui` first — `node_modules/` is partially tracked here but jest isn't in it.

Thanks Vamshi for the heads up on the #960 conflict.

AnDo081105 (approved):
The current version is conflicted with Vamshi 960 PR. Rebase and preserve #960 first while adding shared-client error handling. I'll put this as approve but rebase it first then merge to main. Thank you very much

AnDo081105:
Please resolve the conflicts for this branch. Thanks

PR #947, opened 2026-08-01
feat(api): add API key auth stub for engine ingestion endpoint

jayden-s222587117:


Maneesh-Kumar-Maddini (commented):
Current main already contains engine API-key authentication, while this PR introduces a separate configuration name and key contract. Please compare the original requirement with the existing implementation and close this PR if the requirement is already satisfied. If a gap remains, submit only that remaining change using the established authentication contract, with the real engine caller and credential tests updated consistently. Do not merge this implementation unchanged on top of the existing one.

PR #948, opened 2026-08-01
EE/KF/main engine tflite integration

KishanFW:
- Added EfficientNetV2 TFLite support to the production Engine.
- Added model loading, preprocessing, class mapping, and tensor validation.
- Enabled inference for simulator, Recording Mode, and raw IoT audio.
- Confirmed successful Docker and MQTT inference.
- Fixed the Backend Internal Server Error that prevented Engine predictions from being stored in MongoDB.
- Identified that confidence was validated as a Python Decimal, which PyMongo could not encode.
- Updated the Backend event schema to use a validated float confidence value.
- Confirmed successful end-to-end storage by receiving created event records with MongoDB _id values.


Mangon3 (approved):
Everything looks good to me.

PR #949, opened 2026-08-02
feat(api): restrict CORS to HMI origins instead of wildcard

jayden-s222587117:


Maneesh-Kumar-Maddini (commented):
The origin allowlist is implemented, but this branch is substantially behind main and needs reconciliation with the existing central CORS configuration. Please use one configuration source and verify actual HMI requests, authenticated preflights, allowed origins, and rejected origins. Account for browser headers required by any integrated ETag/correlation functionality. Attach results from the updated application before approval; the old branch’s successful build does not establish current HMI compatibility.

PR #953, opened 2026-08-03
EHMI/BG/fr-b2-audio-recording

BalasaitejaG:
Task- FR-B2: Audio Recording Workflow for the HMI microphone panel.

Operators can now record live browser audio from a selected microphone, stop or cancel cleanly, and replay the latest clip. The flow is simple, reliable, and matches the Sprint 1 HMI requirements.

### What changed
- Added clear **Record / Stop / Cancel** controls on the mic panel
- Added a live recording timer with a **20-second max duration**
- Added a simple green **Play** button to replay the most recent recording
- Improved microphone permission and recording error messages with user-friendly toasts
- Ensured **Stop does not autoplay or trigger unexpected downloads**
- Cleaned up the recording UI so it is easier to use during demo/review

### Files updated
- `src/production/hmi/ui/public/js/audio_recorder.js`
- `src/production/hmi/ui/public/js/HMI.js`
- `src/production/hmi/ui/public/index.html`
- `src/production/hmi/ui/public/css/HMI_style.css`

## Why this matters
FR-B2 is the foundation of the Role B audio UI work. It gives us a working browser recording path before later tasks connect clips to detections (FR-B3) and analysis tools (FR-B4).

## How to test
1. Open the HMI map and click **Click to start**
2. Click a **microphone marker**
3. Press **Record**, allow mic access, speak briefly, then **Stop**
4. Press the green **Play** button and confirm playback
5. Start recording again and press **Cancel** to confirm it discards the take
6. Deny mic permission once and confirm a clear error message appears
7. Let a recording run to about **20 seconds** and confirm it auto-stops
8. Confirm the map still loads normally

## Out of scope
- FR-B1 Spectrogram rendering
- FR-B3 Event audio playback integration
- FR-B4 Audio analysis controls
- Saving recordings to backend/disk

## Acceptance criteria covered
- [x] Start, stop, cancel, and replay through explicit UI controls
- [x] Friendly microphone permission error messages
- [x] No unexpected download/autoplay on stop
- [x] Recording duration shown and capped

### DIFF
--- a/src/production/hmi/ui/public/js/audio_recorder.js
+++ b/src/production/hmi/ui/public/js/audio_recorder.js
@@ -6,17 +6,21 @@
  * stop() returns { blob, samples, sampleRate, mimeType }
  */
 export function getAudioRecorder() {
+  const MAX_RECORDING_SECONDS = 20;
+
   const audioRecorder = {
     audioBlobs: [],
     streamBeingCaptured: null,
     discardChunks: false,
     isRecording: false,
+    isStarting: false,
     audioContext: null,
     scriptProcessor: null,
     mediaStreamSource: null,
     muteGain: null,
     mediaRecorder: null,
     pcmChunks: [],
+    ownedStreams: new Set(),
     sampleRate: 44100,
     recorderMimeType: "audio/webm",
 
@@ -29,11 +33,15 @@ export function getAudioRecorder() {
         );
       }
 
-      if (audioRecorder.isRecording) {
+      if (audioRecorder.isRecording || audioRecorder.isStarting) {
         return Promise.reject(new Error("Recording already in progress"));
       }
 
+      // Lock immediately to prevent concurrent getUserMedia() calls.
+      audioRecorder.isStarting = true;
+
       return navigator.mediaDevices.getUserMedia({ audio: true }).then((stream) => {
+        audioRecorder.ownedStreams.add(stream);
         audioRecorder.streamBeingCaptured = stream;
         audioRecorder.audioBlobs = [];
         audioRecorder.pcmChunks = [];
@@ -93,6 +101,7 @@ export function getAudioRecorder() {
         const finishStart = () => {
           setupPcmGraph();
           audioRecorder.isRecording = true;
+          audioRecorder.isStarting = false;
         };
 
         if (ctx.state === "suspended") {
@@ -100,12 +109,20 @@ export function getAudioRecorder() {
         }
         finishStart();
       }).catch((err) => {
+        audioRecorder.isStarting = false;
         audioRecorder.isRecording = false;
         audioRecorder.cleanupGraph();
-        audioRecorder.stopStream();
+        audioRecorder.stopAllOwnedStreams();
         audioRecorder.resetRecordingProperties();
         throw err;
       });
+    },
+
+    clampSamplesToMax: function (samples, sampleRate) {
+      if (!samples || samples.length === 0) return samples;
+      const maxSamples = Math.floor(sampleRate * MAX_RECORDING_SECONDS);
+      if (samples.length <= maxSamples) return samples;
+      return samples.slice(0, maxSamples);
     },
 
     mergePcmChunks: function () {
@@ -179,6 +196,11 @@ export function getAudioRecorder() {
 
     stop: function () {
       return new Promise((resolve, reject) => {
+        if (audioRecorder.isStarting) {
+          reject(new Error("Recording is still starting"));
+          return;
+        }
+
         if (!audioRecorder.isRecording) {
           reject(new Error("No active recording"));
           return;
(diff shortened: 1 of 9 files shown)


AnDo081105 (changes requested):
## Request changes

1. **P1: Prevent concurrent recording starts from leaking a microphone stream.**

   In `src/production/hmi/ui/public/js/audio_recorder.js:32`, `isRecording` remains false while `getUserMedia()` is pending, and the Record button is not disabled until acquisition succeeds. Two rapid Record clicks can therefore acquire two streams, with the second overwriting the first reference. Stop then releases only the second stream; the mocked reproduction returned `{"stopped":[0,1]}`. Please lock the start operation synchronously before requesting permission and ensure every acquired stream remains owned and cleanable.

2. **P2: Enforce the 20-second maximum as a hard output cap.**

   In `src/production/hmi/ui/public/js/HMI.js:1804`, the maximum duration depends on a UI `setInterval` calling Stop. Background-tab timer throttling or a blocked main thread can delay that callback, while the recorded PCM is never truncated. The resulting clip can exceed 20 seconds. Please retain the timer for UI behavior but also clamp the captured/output samples to `MAX_RECORDING_SECONDS`.

The explicit controls, friendly permission messages, and removal of autoplay/download from the Stop path are in place, but these two lifecycle issues need to be resolved before the acceptance criteria are fully met.

BalasaitejaG:
**P1 fixed: concurrent Record starts no longer leak streams**
Updated src/production/hmi/ui/public/js/audio_recorder.js:

Added a synchronous start lock with isStarting set before getUserMedia()
start() now rejects when either isRecording or isStarting is true
Added ownedStreams tracking (Set) so every acquired stream is owned
Added stopAllOwnedStreams() and now call it in:
start() error cleanup
stop() finalize cleanup
cancel()
Result: rapid repeated starts cannot create orphan microphone streams
Also in src/production/hmi/ui/public/js/HMI.js:

Record button is now disabled immediately while permission is pending (before acquisition succeeds), preventing UI double-start races.


**P2 fixed: 20s limit now hard-clamped in output samples**
Implemented hard cap independent of UI timer:

In audio_recorder.js:

Added MAX_RECORDING_SECONDS = 20
Added clampSamplesToMax(samples, sampleRate)
In stop(), merged PCM samples are clamped before WAV encoding and before returning result.
In HMI.js:

Added defensive clampRecordingSamples(...) and clamp in prepareBrowserRecordingPlayback(...) so replay path also cannot exceed cap even if upstream behavior changes.
Timer UI still exists (for user feedback), but output is now capped at sample level.

src/production/hmi/ui/public/js/audio_recorder.js
```diff
@@ -1,253 +1,311 @@
...
+            let mimeType = "audio/wav";
+            if (clampedSamples.length > 0) {
+              playbackBlob = audioRecorder.encodeWavBlob(clampedSamples, rate);
+              mimeType = "audio/wav";
+            } else if (audioRecorder.audioBlobs.length > 0) {
+              mimeType =
+                (audioRecorder.mediaRecorder && audioRecorder.mediaRecorder.mimeType) ||
+                audioRecorder.recorderMimeType ||
+                "audio/webm";
+              playbackBlob = new Blob(audioRecorder.audioBlobs, { type: mimeType });
```
AnDo081105: **P2 still needs one fix**

- **Problem:** This fallback combines all MediaRecorder chunks without applying the 20-second limit.
- **When it happens:** The fallback is used when no PCM samples are available.
- **Why it matters:** If the timer is delayed, the returned recording can be longer than 20 seconds.
- **Suggested fix:** Combine the MediaRecorder chunks, then decode the blob into audio samples.
- **Apply the limit:** Keep only the first sampleRate times MAX_RECORDING_SECONDS samples.
- **Return a safe clip:** Encode the limited samples as a WAV blob and return that blob.
- **Handle errors:** If the fallback blob cannot be decoded, show a friendly error instead of returning an uncapped recording.
- **Add tests:** Confirm both the normal PCM path and the no-PCM fallback always return recordings of 20 seconds or less.

AnDo081105 (changes requested):
P1 is resolved: concurrent starts are now locked and acquired streams are cleaned up. One P2 issue remains in the MediaRecorder fallback, noted inline. Requesting changes until every returned recording path is capped at 20 seconds.

AnDo081105 (approved):
Re-review of dcfe895c confirms the remaining P2 issue is resolved.

- The no-PCM MediaRecorder fallback now combines and decodes the captured blob, clamps decoded samples to sampleRate * 20, and re-encodes the capped result as WAV.
- Decode failures return a friendly processing error instead of returning an uncapped blob.
- Added tests cover the PCM path, the no-PCM fallback at 48 kHz, and fallback decode failure; all 5 tests passed in verification.
- P1 remains resolved.

No blocking findings remain within the agreed review scope.

PR #960, opened 2026-08-05
FR-A1: Unified Live Map Rendering

Vamshi-Gollapelly:
## Summary

Implements FR-A1: Unified Live Map Rendering for the EchoNet HMI.

## Changes

- Uses Leaflet as the single production map-rendering flow.
- Prevents duplicate Leaflet map initialisation using a singleton guard.
- Retrieves IoT nodes through the shared `retrieveIotNodes()` API helper.
- Prevents duplicate markers and connection lines during repeated refreshes.
- Uses shared HMI utilities for loading states, retry handling, error messages, and HTML escaping.
- Documents the selected production map architecture in `map.js`.

## Testing

- Verified the map loads successfully at `http://localhost:3000/map`.
- Confirmed repeated page refreshes do not create duplicate map instances.
- Confirmed markers and connection lines do not accumulate.
- Confirmed no FR-A1-related JavaScript or Leaflet errors occur during testing.


src/production/hmi/ui/public/js/map.js
```diff
@@ -48,145 +45,222 @@ const childIcon = L.icon({
...
-// ─────────────────────────────────────────────────────────────────────────────
-// Map initialisation
-// ─────────────────────────────────────────────────────────────────────────────
-
+/**
+ * Initialise the canonical Leaflet map once.
+ */
 async function initMap() {
-  map = L.map("map").setView([-38.7789, 143.5705], 14);
+  const mapContainer = document.getElementById("map");
```
AnDo081105: The production page contains `#basemap`, not `#map`, and it never imports this module, so this initializer is not executed. Please apply this logic to the active OpenLayers implementation in `nodes-overlay.js`. The core structure should be:

```js
import { retrieveIotNodes } from "./routes.js";

async function addIoTNodesToMap(hmiState) {
  const response = await retrieveIotNodes();
  const nodes = Array.isArray(response.data) ? response.data : [];

  if (!hmiState.iotNodeLayer) {
    hmiState.iotNodeLayer = createIotNodeLayer();
    hmiState.basemap.addLayer(hmiState.iotNodeLayer);
  }

  const source = hmiState.iotNodeLayer.getSource();
  source.clear();

  const validNodes = nodes.filter(hasValidCoordinates);
  addNodeFeatures(source, validNodes);
  addUniqueConnectionFeatures(source, validNodes);
  initialiseNodePopupOnce(hmiState);
}
```

This preserves the existing detection and microphone layers while ensuring repeated node refreshes do not accumulate layers, markers, lines, overlays, or listeners.

AnDo081105 (commented):
The new Leaflet map code is not connected to the page users actually open. `/map` still serves `index.html`, which loads OpenLayers through `HMI.js` and renders into `#basemap`. It does not import `map.js` or load Leaflet. As a result, this PR does not change the running map, and the existing node overlay can still add duplicate layers and connection lines.

Keep OpenLayers as the single production renderer and move the shared API, layer-reuse, coordinate-validation, and connection-deduplication logic into `nodes-overlay.js`, which is the active path called by `HMI.js`. Automated coverage should confirm that repeated node loads retain one layer and one line per logical connection.

src/production/hmi/ui/public/js/nodes-overlay.js
```diff
@@ -1,135 +1,293 @@
...
+  try {
+    const response = await retrieveIotNodes();
+
+    const nodes = Array.isArray(response?.data)
+      ? response.data
+      : [];
+
+    const validNodes = nodes.filter(hasValidCoordinates);
+
+    if (!hmiState.iotNodeLayer) {
```
AnDo081105, line 265: This only reuses the IoT layer. It does not stop the OpenLayers map itself from being created again. `initialiseHMI()` always calls `createBasemap()`, and the retry button calls `initialiseHMI()` again. Please make map initialisation idempotent by reusing or disposing the existing map, and add a repeated-initialisation test.

src/production/hmi/ui/public/js/nodes-overlay.js
```diff
@@ -1,135 +1,293 @@
...
+     */
+    source.clear();
+
+    addNodeFeatures(source, validNodes);
+    addUniqueConnectionFeatures(source, validNodes);
+
+    initialiseNodePopup(hmiState);
+  } catch (error) {
+    console.error("Error loading IoT nodes:", error);
+    throw error;
```
AnDo081105, line 285: `addIoTNodesToMap()` throws here, but its caller in `HMI.js` does not await or return the promise. The page can show "Map data loaded successfully" while this request fails, and the rejection is unhandled. Please await or return this call and handle the failure before showing success.

Vamshi-Gollapelly:
Fixed the map one by making createBasemap() reuse hmiState.basemap if it's
already there instead of spinning up a second ol.Map every time the retry
button gets hit. While I was in there I also gated the rest of
initialiseHMI()'s one-time setup (layers, click handler, all of that)
behind the same check — turns out that stuff wasn't guarded either, so a
retry would've quietly stacked a second copy of every layer on top too.

Actually turned up something extra while chasing this down: since the map
used to get rebuilt on every retry, the IoT layer/popup guards you'd have
hit next were pointing at the *old*, now-orphaned map. So nodes would've
just silently stopped showing up after a retry. Reusing the map fixes that
side effect as well, so it was worth catching now rather than in a
follow-up round.

For the async one — the mic-fetch .then() now awaits addIoTNodesToMap()
before it does anything else, so if that call fails it falls into the
existing catch/retry/toast path instead of leaving an unhandled rejection
while the page tells you everything loaded fine.

Also swapped the test file over to actually import and run the real
functions from nodes-overlay.js rather than a reimplementation of the same
logic — 12 tests now, node --test, including the repeated-refresh and
repeated-init cases you'd want covered here.

src/production/hmi/ui/public/js/nodes-overlay.js
```diff
@@ -1,135 +1,293 @@
...
+ * The IoT vector layer, popup overlay and pointer listener are created once
+ * and reused. The vector source is cleared before every node refresh so
+ * repeated updates cannot accumulate duplicate markers or connection lines.
+ */
+
+import { retrieveIotNodes } from "./routes.js";
+
+const IOT_LAYER_NAME = "iot_nodes";
+
+function hasValidCoordinates(node) {
```
AnDo081105, line 23: Small comment, `hasValidCoordinates()` converts values before checking them. JavaScript converts null and "" to 0, so invalid nodes can appear at the wrong location. Reject null and empty values before conversion

AnDo081105 (approved):

AnDo081105 (approved):
Please resolve the conflicts

PR #961, opened 2026-08-05
EH/KIM/fr-a2-mqtt-event-handling

kimmynoto:
## Summary
This implements FR-A2 (Live MQTT Event Handling). There wasn't actually any MQTT connection in the codebase before this just a placeholder variable waiting for a handler that never got built. So this PR builds that whole piece from scratch such as connecting to the broker, tracking connection state, cleaning up incoming data, and letting users know when the connection drops or comes back.

## What's in this PR
- Built a new MQTT client (mqtt_client.py) that connects to the broker and subscribes to live event topics
- Hooked it into the backend's startup, with error handling so a failed connection just retries quietly in the background instead of crashing the whole API
- Added automatic reconnect with backoff
- Added a /mqtt/connection-state endpoint so the connection status is actually checkable
- Added normalization for incoming vocalization events so they come out in one consistent shape instead of raw/inconsistent MQTT payloads
- Added a frontend piece that polls the connection state and shows a banner/toast (using the existing showPageBanner/showToast helpers) when the connection drops or recovers

## How I tested it
- Confirmed the MQTT client connects successfully via the container logs
- Manually published a test message with mosquitto_pub and confirmed it comes out normalized correctly
- Stopped and restarted the MQTT broker container to confirm the frontend banner/toast actually show up and clear correctly


src/production/backend/app/services/mqtt_client.py
```diff
@@ -0,0 +1,78 @@
...
+connection_state = "disconnected"  # disconnected | connecting | connected | reconnecting
+
+MQTT_BROKER_URL = os.environ.get("MQTT_BROKER_URL", "ts-mqtt-server-cont")
+MQTT_BROKER_PORT = int(os.environ.get("MQTT_BROKER_PORT", 1883))
+
+def on_connect(client, userdata, flags, rc):
+    global connection_state
+    connection_state = "connected"
+    print(f"[MQTT] Connected, rc={rc}")
+    client.subscribe("Simulate_Recording")
```
AnDo081105: [P1] This subscribes to `Simulate_Recording`, but the simulator publishes vocalization events to `MQTT_PUBLISH_URL` (`projectecho/engine/2`). This client will miss normal simulator events. Please subscribe to the actual live-event topic, preferably from configuration.

src/production/backend/app/services/mqtt_client.py
```diff
@@ -0,0 +1,78 @@
...
+def on_disconnect(client, userdata, rc):
+    global connection_state
+    connection_state = "reconnecting"
+    print(f"[MQTT] Disconnected, rc={rc} - attempting reconnect")
+
+def on_message(client, userdata, msg):
+    normalized = normalize_payload(msg.payload, msg.topic)
+    if normalized:
+        print(f"[MQTT] Normalized event: {normalized}")
+    # TODO: forward `normalized` to wherever the frontend/dashboard reads from
```
AnDo081105, line 33: [P1] The normalized event stops here and is only logged. Nothing sends or stores it for the HMI, so no live event can appear on the map or dashboard. Please connect this output to the frontend's live-event transport or polling source.

src/production/hmi/ui/public/js/HMI.js
```diff
@@ -349,8 +349,51 @@ fetch("./js/sample_data.json")
...
+      } else if (state === "reconnecting") {
+        showPageBanner("Live data connection lost — reconnecting…", "warning", false);
+        showToast("Live data connection lost, reconnecting…", "warning");
+      } else if (state === "disconnected") {
+        showPageBanner("Live data unavailable", "error", false);
+      }
+      _lastMqttState = state;
+    }
+  } catch (err) {
+    console.error("Error polling MQTT connection state:", err);
```
AnDo081105, line 623: [P1] If this request fails, the page only logs the error and can continue showing the last `connected` state. Please set an unavailable or fallback state and show the user-facing banner here.

AnDo081105 (commented):
P1 findings from the FR-A2 review.

src/production/backend/app/main.py
```diff
@@ -60,6 +60,22 @@
...
+@app.on_event("startup")
+def startup_mqtt():
+    start_mqtt_client()
+
+@app.get("/mqtt/connection-state", tags=["mqtt"])
+def mqtt_connection_state():
+    return {"state": get_connection_state()}
+
+@app.get("/mqtt/latest-events", tags=["mqtt"])
+def mqtt_latest_events():
```
AnDo081105, line 75: This endpoint stores and returns events, but `HMI.js` never calls it, so the events still do not reach the map or dashboard. Please add frontend polling or a WebSocket consumer and pass new events to `updateVocalizationLayerFromLiveData`. The response must also include the fields expected by `convertJSONtoAnimalVocalizationEvent`, such as species, type, status, diet, location, confidence, and a stable event ID.

AnDo081105 (commented):
Additional FR-A2 review findings.

src/production/backend/app/services/mqtt_client.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    try:
+        data = json.loads(raw_payload)
+    except (json.JSONDecodeError, TypeError):
+        print(f"[MQTT] Could not parse payload on {topic} as JSON")
+        return None
+
+    # Vocalization / recording events (from comms_manager.py's
+    # mqtt_send_random_audio_msg and mqtt_send_recording_msg)
+    if "animalEstLLA" in data or "audioClip" in data:
```
AnDo081105, line 52: FR-A2 also requires movement, sensor-health, and IoT-node updates. This function only handles vocalizations and marks every other payload as `unknown`. Please add a normalized shape for each required event type and route each type to the correct HMI update function.

src/production/backend/app/main.py
```diff
@@ -60,6 +60,24 @@
...
+    from app.routers.hmi import show_latest_events
+    return {"events": show_latest_events(limit=20)}
```
AnDo081105:
```suggestion
@app.get("/mqtt/latest-events", tags=["mqtt"])
def mqtt_latest_events():
    from app.routers.hmi import show_latest_events

    events = [
        {"eventType": "vocalization", **event}
        for event in show_latest_events(limit=20)
    ]
    events += [
        event for event in get_latest_events()
        if event.get("eventType") != "vocalization"
    ]

    return {"events": events}
```

src/production/backend/app/services/mqtt_client.py
```diff
@@ -0,0 +1,123 @@
+import paho.mqtt.client as mqtt
+import os
+import json
+
+connection_state = "disconnected"  # disconnected | connecting | connected | reconnecting
+latest_events = {}
+
+MQTT_BROKER_URL = os.environ.get("MQTT_BROKER_URL", "ts-mqtt-server-cont")
+MQTT_BROKER_PORT = int(os.environ.get("MQTT_BROKER_PORT", 1883))
+MQTT_TOPIC = os.environ.get("MQTT_PUBLISH_URL", "projectecho/engine/2")
```
AnDo081105:
```suggestion
MQTT_TOPICS = os.environ.get(
    "MQTT_TOPICS",
    "projectecho/engine/2,projectecho/movement,iot/data/test",
).split(",")


def on_connect(client, userdata, flags, rc):
    global connection_state
    connection_state = "connected"

    for topic in MQTT_TOPICS:
        client.subscribe(topic.strip())
```

AnDo081105:
Code suggestion: publish movement events to the MQTT feed.

GitHub cannot attach this inline because `src/production/simulator/src/comms_manager.py` is not changed by this PR. Insert after `movement_event` is constructed (currently line 165):

    self.mqtt_client.publish(
        os.environ.get("MQTT_MOVEMENT_TOPIC", "projectecho/movement"),
        json.dumps(movement_event),
    )

This keeps the existing HTTP persistence and adds the MQTT-derived movement event required by FR-A2.

src/production/hmi/ui/public/js/HMI.js
```diff
@@ -589,8 +589,110 @@ fetch("./js/sample_data.json")
...
+      _seenMqttEventIds.add(event._id);
+
+      if (event.eventType === "vocalization") {
+        newVocalizationEvents.push(event);
+      } else if (event.eventType === "movement") {
+        newMovementEvents.push(event);
+      }
+      // sensor_health / iot_node: normalized on the backend, but there is
+      // no existing HMI render function for these yet, so they are not
+      // routed to the map here. Flagged for a follow-up ticket.
```
AnDo081105: Route every normalized MQTT event type. The current branch handles only vocalization and movement, then explicitly drops sensor-health and IoT-node updates required by FR-A2.

Replace this routing block:

```suggestion
switch (event.eventType) {
  case "vocalization":
    newVocalizationEvents.push(event);
    break;
  case "movement":
    newMovementEvents.push(event);
    break;
  case "sensor_health":
  case "iot_node":
    document.dispatchEvent(
      new CustomEvent(`mqtt:${event.eventType}`, { detail: event })
    );
}
```

AnDo081105 (commented):
I have made some suggestions based on my findings. Have a look and re-check to see whether I missed anything. Thank you so much:D

AnDo081105 (approved):

src/production/hmi/ui/public/js/HMI.js
```diff
@@ -12,7 +12,8 @@
  *              No changes to map logic, layer management, or audio handling.
  */
 
-import { showToast, getApiErrorMessage, withRetry } from "./HMI-utils.js";
+import { showToast, getApiErrorMessage, withRetry, showPageBanner, hidePageBanner } from "./HMI-utils.js";
+// TEMP: audio_recorder.js has a pre-existing syntax error, commented out to test FR-A4 locally
 import { getAudioRecorder } from "./audio_recorder.js";
```
Vamshi-Gollapelly, line 16: This whole commit (Start FR-A4: Live Map Controls and Filters) landed after Khanh's approval, so it hasn't actually been reviewed, and it's out of scope for this PR, this one's about FR-A2/MQTT.
Bigger issue though: you're re-enabling this import with a comment saying audio_recorder.js has a pre-existing syntax error. If that's accurate, importing it will throw and could break the whole HMI page on load, not just the new map controls.

Can we split the FR-A4 map controls work into its own PR, and either fix the syntax error in audio_recorder.js first or leave this import commented out until it's actually fixed? Don't think this should go in as part of the MQTT work.

src/production/hmi/ui/public/js/HMI.js
```diff
@@ -589,11 +591,147 @@ fetch("./js/sample_data.json")
...
  */
+// ─────────────────────────────────────────────────────────────────────────────
+// MQTT connection state polling (FR-A2)
+// ─────────────────────────────────────────────────────────────────────────────
+
+let _lastMqttState = null;
+
+async function pollMqttConnectionState() {
+  try {
+    const response = await fetch("http://localhost:9000/mqtt/connection-state");
```
Vamshi-Gollapelly, line 601: This is hardcoded to http://localhost:9000, same thing a couple lines down in pollMqttLatestEvents for /mqtt/latest-events. Works fine for the local docker-compose setup but will just fail for anyone running this anywhere else, since there's no way to point it at a different host.

Might be worth pulling this from a config value or the same base URL the rest of the app's API calls already use, rather than hardcoding it twice here.

src/production/backend/app/services/mqtt_client.py
```diff
@@ -0,0 +1,133 @@
...
+    global connection_state
+    connection_state = "reconnecting"
+    print(f"[MQTT] Disconnected, rc={rc} - attempting reconnect")
+
+def on_message(client, userdata, msg):
+    normalized = normalize_payload(msg.payload, msg.topic)
+    if normalized:
+        print(f"[MQTT] Normalized event: {normalized}")
+        key = normalized.get("_id", "unknown")
+        latest_events[key] = normalized
```
Vamshi-Gollapelly, line 32: This dict just keeps growing, every message that comes in adds a new key and nothing ever removes one. Over a long-running deployment that's a slow memory leak on the backend, and /mqtt/latest-events will keep returning a bigger response every time since it's returning everything ever seen, not just recent stuff.

Also, this write happens on the MQTT client's own thread (from loop_start()), while the /mqtt/latest-events handler reads it from a different thread with no lock in between. list(latest_events.values()) while this line is mutating the dict on another thread could occasionally throw a "dictionary changed size during iteration" error.

Would suggest capping this (e.g. a bounded deque or evicting old entries) and adding a lock around the read/write, similar to what's already being done for connection_state elsewhere maybe not, but worth double checking that one too.

Vamshi-Gollapelly (dismissed):
Also noting: no CI checks ran on this PR at all (Checks shows 0), and there's a merge conflict against main now too, will need a rebase before this can go in either way.

The MQTT normalization/reconnect work itself (the stuff Khanh already went through three rounds on) looks solid. Main blocker for me is the FR-A4 commit that snuck in after approval, that needs to come out or go in its own PR, especially given the syntax error note. The localhost URL and the events dict growth are smaller but worth fixing while we're in here.

KishanFW (approved):

PR #967, opened 2026-08-06
docs(architecture): add backend dependency map for Engine and MongoDB

jayden-s222587117:


KCP17 (approved):
Well-verified doc. However, line-number citations like engine.py:17-25 only stay correct for as long as main doesn't change. Since main on this repo changes constantly, those line numbers will eventually point at the wrong code. I'll still approve this for now since we're at the end of the project, but next time please consider using a GitHub permalink instead (click on the 3 dots of a specific code line -> "Copy permalink") to lock the link to the exact version of the file at that moment, so it stays accurate forever, no matter how much main changes later.

PR #973, opened 2026-08-08
feat(backend): global configuration, exception handling and security headers

jayavardhan438-cell:
Covers all four checklist items of the Planner task **Global Configurations and Security Headers** (B4.2, C1.5, C4.3, security header middleware).

Three commits, eleven files, 75 tests. No new runtime dependency.

## What's included

| Item | Change |
|---|---|
| B4.2 | Audit only — no code. Seven hardcoded lines across three files, grouped into five findings. Documented separately. |
| C1.5 | New `app/config.py`: one place where environment settings are read. Required secrets raise `ConfigError` when unset instead of falling back to a credential written into source. 23 tests. |
| C4.3 | New `app/error_handlers.py`: application-wide handlers for unhandled exceptions, `HTTPException` and validation errors. Unhandled errors return a generic message and a `request_id`; the full traceback goes to the server log only. 14 tests. |
| Security headers | New `app/middleware/security_headers.py`: `X-Content-Type-Options`, `X-Frame-Options`, `Referrer-Policy`, `Cross-Origin-Opener-Policy`, `X-Permitted-Cross-Domain-Policies`, `Permissions-Policy` and `Content-Security-Policy` on every response. 38 tests. |

## Testing

```
pytest src/tests/unit/backend -q
75 passed
```

Two of the config tests are regression tests for audit findings F-01 and F-02. They exist to fail if a hardcoded credential is ever reintroduced as an `os.getenv` default.

## Verified against the running container

Error handling — request to a path that does not exist:

```
# before
{"detail":"Not Found"}

# after
{"error":{"type":"http_error","message":"Not Found","request_id":"d3a37c74529f"}}
```

Security headers — `curl -s -D - -o /dev/null http://localhost:9000/` returns all seven headers, and no `server:` header. `/docs` was opened in a browser and renders correctly under the Content-Security-Policy.

## Deliberately not included

**`config.py` is not yet wired into `database.py`, `main.py` or `hmi.py`.** The hardcoded values found in the audit are unchanged in those files. Please don't read this PR as resolving them.

This was separated on purpose. A module nothing imports cannot change the behaviour of the running application, so it can be reviewed on its own. Removing the credential fallbacks changes startup for every developer and every container — an environment that doesn't supply `MONGODB_URI` will refuse to start, by design — and that needs agreement first on how the variables are supplied in each environment.

**CORS is unchanged.** `main.py` currently sets `allow_origins=["*"]` together with `allow_credentials=True`. The CORS specification forbids that combination and browsers reject it, so the configuration doesn't behave as intended. `config.cors_allowed_origins()` exists and is tested, so the fix is small — but the correct list of origins isn't mine to choose. The HMI container publishes port 3000 while the existing hardcoded value refers to 8080. Happy to do it in this PR if you tell me which origins are right, or leave it for a separate change.

## Notes for the reviewer

**Where the middleware is registered matters.** `main.py` constructs `FastAPI()` twice, at lines 22 and 39; the first instance and everything attached to it is discarded when `app` is reassigned. Both registrations are placed after line 39, with a comment explaining why. Anything attached before that line silently has no effect.

**Two things I found but did not change**, since fixing them would enlarge this PR into unrelated territory:

- The duplicate `FastAPI()` construction above should probably be removed outright.
- Because of it, `app.include_router(projects.router)` on line 34 attaches to the discarded instance — so every endpoint in `app/routers/projects.py` is unreachable in the running API. This looks unnoticed and is worth a ticket.

**`--no-server-header`** was added to `API.Dockerfile` and the docker-compose command. uvicorn writes `server: uvicorn` after the ASGI middleware stack has run, so middleware cannot remove it. It's needed in both places because the compose `command:` overrides the image `CMD`.

**Line endings.** Files were authored on Windows; `git diff` should show additions only. No existing line of application code was modified apart from the two uvicorn command lines.

## Related

Branch `EB/AVD/Backend-config-logging-linter` appears to cover related ground. I haven't found any file overlap with this change, but worth a look before merging.

The failing `Docker Image CI / build` check is pre-existing and not caused by this branch.

`app/routers/sim.py` line 15 constructs `ConnectionConfig` at import time with
`MAIL_STARTTLS` and `MAIL_SSL_TLS` commented out (lines 21-22), so fastapi-mail
falls back to reading them from the environment. `docker-compose.yml` supplies
both at lines 111-112; `docker-compose.test.yml`, which CI builds, supplies
neither. The API container therefore fails to start and the job's nine-container
check fails.

Reproduced locally on the test compose:

    pydantic.error_wrappers.ValidationError: 2 validation errors for ConnectionConfig
    MAIL_STARTTLS
      field required (type=value_error.missing)
    MAIL_SSL_TLS
      field required (type=value_error.missing)

No file in this PR is in that code path. The fix is two lines in
`docker-compose.test.yml` mirroring the main compose file — happy to include it
here if you'd prefer, or raise it separately.

### DIFF
--- a/src/deployment/docker/docker-compose.yml
+++ b/src/deployment/docker/docker-compose.yml
@@ -114,7 +114,10 @@ services:
 
     stdin_open: false
     tty: true
-    command: uvicorn app.main:app --host 0.0.0.0 --port 9000 --reload
+    # --no-server-header suppresses uvicorn's "server: uvicorn" response
+    # header. It must be set here as well as in API.Dockerfile, because this
+    # command overrides the image's CMD.
+    command: uvicorn app.main:app --host 0.0.0.0 --port 9000 --reload --no-server-header
 
   mongo-express:
     image: mongo-express
--- a/src/production/backend/API.Dockerfile
+++ b/src/production/backend/API.Dockerfile
@@ -41,4 +41,7 @@ COPY . /app
 EXPOSE 9000
 
 # Start the API using uvicorn
-CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "9000"]
+# --no-server-header suppresses uvicorn's "server: uvicorn" response header,
+# which discloses the server software. It cannot be removed by ASGI
+# middleware because uvicorn writes it after the middleware stack has run.
+CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "9000", "--no-server-header"]
--- a/src/production/backend/app/config.py
+++ b/src/production/backend/app/config.py
@@ -74,6 +74,34 @@ def get_list(name, default):
     return items or list(default)
 
 
+_TRUE_VALUES = ("1", "true", "yes", "on")
+_FALSE_VALUES = ("0", "false", "no", "off")
+
+
+def get_bool(name, default):
+    """Return ``name`` parsed as a boolean, or ``default`` if it is not set.
+
+    Environment variables are always strings, so "false" would be truthy under
+    Python's normal rules. Accepted spellings are listed explicitly and
+    anything else raises, because silently treating an unrecognised value as
+    False is how a security setting ends up switched off without anyone
+    noticing.
+    """
+    value = _read(name)
+    if value is None:
+        return default
+    lowered = value.lower()
+    if lowered in _TRUE_VALUES:
+        return True
+    if lowered in _FALSE_VALUES:
+        return False
+    raise ConfigError(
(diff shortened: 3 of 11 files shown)


Anand152002 (approved):
Reviewed the changes against the four checklist items. The hardcoded URL/port audit, environment variable fallback tests, global exception handlers, and security header middleware are all covered with appropriate test coverage. The Uvicorn Server header handling is also correctly addressed in both the Dockerfile and Compose configuration.

Looks good to me.

Anand152002 (approved):

PR #974, opened 2026-08-08
Sprint 1 inference wrapper prototype

Anand152002:
## Overview

This pull request contains the Sprint 1 prototype implementation for the Project Echo Engine inference wrapper.

The implementation was developed exclusively within the prototype environment and does not modify the production Engine, Backend, HMI, or production Docker configuration.

## Sprint 1 Objectives Completed

- Finalised the proposed standard Engine inference response.
- Preserved the existing simulator and prediction fields.
- Implemented a lightweight Engine-side inference wrapper in the prototype environment.
- Used the existing prediction result and simulator metadata to generate the standard payload.
- Added validation for normal and failure scenarios.
- Tested compatibility with the current Backend payload requirements.
- Identified schema differences between the proposed response and the current Backend.
- Documented required future changes for the Engine, Backend, and HMI teams.

## Prototype Components

The following files were added:

- `response_schema.py` – Defines the proposed standard inference response.
- `inference_wrapper.py` – Generates standard success and failure responses.
- `validator.py` – Validates required fields, audio input and inference results.
- `backend_adapter.py` – Converts the proposed response into the current Backend payload format.
- `prototype_engine.py` – Coordinates the prototype inference workflow.
- `test_wrapper.py` – Executes the validation scenarios.

## Validation Scenarios

The prototype was tested against the following scenarios:

1. Valid input
2. Missing required fields
3. Invalid audio
4. Inference failure

All four scenarios produced the expected validation behaviour.

## Backend Compatibility

A Backend adapter was used to test compatibility with the existing Backend without modifying the production Backend.

The main schema differences identified were:

- The proposed response introduces a `status` field.
- The proposed response introduces a structured `error` field.
- The proposed schema represents GPS information as structured objects, while the current Backend expects GPS coordinates in array form.

These differences have been documented for future Engine and Backend integration.

## Production Safety

No production Engine, Backend, HMI, or production Docker Compose files were modified as part of this implementation.

All prototype work is located under:

`src/prototypes/engine/Inference_wrapper/`


Mangon3 (approved):
Very clean work!

- Imports are directory-relative (`from response_schema import ...`) so it only runs from inside `Inference_wrapper/`. An `__init__.py` + package-relative imports would let it run/import from anywhere.
- `test_wrapper.py` prints, it doesn't assert (it's a demo, not automated tests). Converting the four scenarios to `pytest` with `assert`s (esp. the failure codes) would make them real.

Other than that everything LGTM!

PR #976, opened 2026-08-08
EE/QDN: Reproducible Model-Training Pipeline Configuration (Sprint 1)

25qdatttt:
## Summary
- I ported the training pipeline code (`main.py`, `train.py`, `dataset.py`, `model/` — 8 files) from `.delete/archive/Prototypes/engine/torch_impl/` into its permanent home at `src/prototypes/engine/augmentation/`, so that Kiernan Nguyen's shared Hydra config has runnable code behind it.
- While porting, I found and fixed 2 pre-existing bugs: a typo in `main.py` (`cfg.run.cKLDivLossheckpoint_path` should be `cfg.run.checkpoint_path`, which was breaking the `run.test=true` checkpoint-loading branch), and a missing `OmegaConf` import in `train.py` that caused a `NameError` in `save_checkpoint`/`load_checkpoint`.
- I seeded `numpy` and `random` alongside `torch` in `main.py`, because augmentation masking uses `random.randint`/`random.random`, so the pipeline wasn't actually reproducible before, only the model init and dataloader shuffling were.
- I rewrote `augment.py`'s `SpecAugment` class, because the archived version's constructor didn't match 4 of the 5 committed augmentation presets (`default`/`light`/`heavy`/`original_unfixed_reference`), so they would all raise a `TypeError` on `hydra.utils.instantiate`. The new version supports both the legacy pixel API and the ratio-based API, and I didn't need to change any preset YAML to make this work.
- I added Kiernan Nguyen's Hydra config (`config.yaml`, 7 `model/` presets, 6 `teacher_model/` presets, 5 `augmentation/` presets, `local/cpu.yaml`) as the shared baseline, with one additive fix on top: I added `norm_choice: freeze_bn` to 4 model presets (`ghost_efficientnet_v2`, `panns_cnn14`, `panns_mobilenetv1`, `panns_mobilenetv2`) that were missing this required key and crashed on model construction. I didn't remove or rename anything.
- I added `config/local/local_data_files.yaml`, a new opt-in local-dev config (following the same pattern as `cpu.yaml`) that points at the real local dataset in `src/prototypes/data_files/`.
- I added a scoped `pyproject.toml` and `uv.lock` (86 packages resolved) for this pipeline's dependencies, kept separate from the repo-root TF-based `requirements.txt` files.
- I added an end-to-end smoke test at `src/tests/pipeline/engine_training/smoke_test/`.
- I added `.gitignore` patterns for pipeline artifacts (`*.pth`, `*.onnx`, `outputs/`, `.cache/`), and also to stop the untracked ~922MB `src/prototypes/data_files/`/`src/prototypes/spectrograms/` local dataset from being accidentally committed.
- I documented all of this in `src/prototypes/engine/augmentation/README.md` (setup and execution instructions, unresolved-dependency list, baseline configuration record, known limitations) and in `docs/training.md` (ported content plus a new SpecAugment section), and I captured a real run in `docs/baseline_smoke_training_log.txt`.

## Test plan
- [x] I ran `python src/tests/pipeline/engine_training/smoke_test/test_train_smoke.py` and got `Ran 1 test in 34.053s — OK`.
- [x] I did a manual run on real data: `ghost_efficientnet_v2`, 3 real species from `src/prototypes/data_files/`, 1 epoch, CPU. It completed successfully and saved a checkpoint.
- [x] I verified all 5 augmentation presets (`config/augmentation/*.yaml`) instantiate via `hydra.utils.instantiate`, and each one ran a dummy tensor through `forward()` without error.
- [x] I stress-tested the `max_total_time_ratio` cumulative-cap behaviour over 500 random trials, and it stayed within the cap every time.
- [x] I confirmed `uv sync` installs cleanly (86 packages) from `src/prototypes/engine/augmentation/pyproject.toml`.

## Notes for reviewers
- I already agreed with Kiernan Nguyen and Praveen via Teams, before opening this PR, about the following: `norm_choice` missing from 4 model preset YAMLs, `augment.py`'s API mismatch with the committed presets, a `val_split` edge case (each class needs at least 5 files, or the validation split rounds down to 0 and crashes `train.py`'s `_evaluate` with a `ZeroDivisionError`), and the seeding/reproducibility gap. I've fixed all of these and detailed them in `README.md` under "Known Limitations".
- I'd like to confirm with the team: is `norm_choice: freeze_bn` the right default for `ghost_efficientnet_v2` and the `panns_*` models? Since `panns_mobilenetv2_qat.yaml` already uses `swap_rms_norm` for its QAT variant, this might not be uniform across every architecture. I made this choice as a judgment call to unblock the pipeline, not as a design decision on the team's behalf, so I'd appreciate a second opinion.
- The 6 `config/teacher_model/*.yaml` presets have the same missing-`norm_choice` gap. I've left this unfixed for now, since the distillation path is deferred to Sprint 2.
- I want to be clear that this pipeline is a new baseline going forward, and not a reproduction of any previously-trained model's actual training run. The repo has 3 different prior "trained models" with undocumented or conflicting provenance (see the README for details), and by team decision, I'm not trying to reverse-engineer any of them.
- I've deferred the QAT preset variants, the `run.test=true` checkpoint-loading path, and a full (non-synthetic) training run against the real ~922MB local dataset to Sprint 2.


Mangon3 (approved):
Looks good, I'll approve but there are some possible improvements:

**`train.py` - bf16 mixed precision looks disabled.** `use_amp = (self.dtype == torch.float16) and "cuda" in ...`, and autocast is `enabled=self.use_amp`. So with the default `dtype: bfloat16`, autocast never turns on and training runs in fp32. bf16 autocast doesn't need the `GradScaler`, so it could be enabled independently of `use_amp`.

**`train.py` - `save_checkpoint`/`load_checkpoint` are unused.** `train()` saves with `torch.save(self.model.state_dict(), ...)` directly and nothing calls `load_checkpoint`, so the richer checkpoint (optimizer/epoch/cfg) and resume path are dead code. Try wiring the code together.

**`quant.py`** Duplicate `import torch`, both `as quant` and `as quantization`, and `numpy` is unused. Also `input_size=(1,3,32,32)` won't match the real spectrogram shape.

PR #978, opened 2026-08-09
EH/OWY/FR-C4: Improve UX Consistency Across Admin Pages

ohwenyi:
**Summary**
Implements FR-C4: UX Consistency Across Admin Pages for Sprint 1.

This PR improves consistency across the Project Echo admin interface by introducing a shared layout approach and improving responsive navigation across production admin pages.

**Changes**
- Added shared `admin-layout.js` for consistent sidebar, header and footer loading.
- Migrated relevant admin pages from repeated page-specific component loading to the shared layout.
- Improved shared sidebar spacing and navigation behaviour.
- Fixed the responsive Dashboard layout so critical header controls remain accessible on mobile screens.
- Improved hamburger/sidebar behaviour across production admin pages.
- Updated Sensor Health subpages, including Alerts, Reboot and Settings, to use the shared layout.
- Resolved page-specific layout conflicts while preserving existing page functionality.

**Testing**
- Tested the relevant admin pages locally using Docker.
- Tested responsive behaviour using Chrome DevTools at 375 × 667.
- Confirmed hamburger/sidebar open and close behaviour.
- Confirmed existing page-specific functionality remained after the shared-layout changes.

**Notes**
Existing placeholder content in `template.html` and `cloud-compute.html` has been retained following confirmation from the team leader.

Ready for review before merging.


src/production/hmi/ui/public/admin/dashboard.css
```diff
@@ -4696,3 +4696,28 @@ perspective effects (not including the modals and the overlay).
   -webkit-animation-name: slideOutUp;
   animation-name: slideOutUp;
 }
+
+/* API status badge */
+.api-status-badge {
+  position: fixed;
+  top: 10px;
```
AnDo081105: The fixed badge overlaps the dashboard header actions at desktop widths. At 1280 × 800, it covers both “Go to Echo” and the profile menu, and hit-testing targets the badge instead of either link. Please place the status message in normal flow below the header, or otherwise reserve space so these controls remain clickable.

src/production/hmi/ui/public/admin/dashboard.html
```diff
@@ -21,21 +22,22 @@
...
   <script type="text/javascript" src="/admin/js/dashboard.js"></script>
   <script type="text/javascript" src="/admin/js/admin-page-state.js"></script>
 </head>
 
 <body>
 
 <!-- Sprint 2: API Status Badge -->
-<div id="api-status-badge" style="position: fixed; top: 10px; right: 10px; z-index: 9999;">Loading status...</div>
-<script src="/js/api-status.js"></script>
+<body>
```
AnDo081105, line 30: This opens a second `<body>` even though the document already opened one on line 29. Browsers may repair the markup differently, and validators will reject it. Please remove this duplicate tag. Also clean the trailing whitespace reported by `git diff --check` in `api-explorer.html` and `notifications.html` before merging.

AnDo081105 (commented):
Please address the two inline comments.

AnDo081105 (changes requested):

src/production/hmi/ui/public/admin/dashboard.css
ohwenyi: Addressed. The API status message is now in normal page flow below the header and no longer overlaps the desktop header controls. I also tested the updated Dashboard locally.

src/production/hmi/ui/public/admin/dashboard.html
ohwenyi: Addressed. I removed the duplicate <body> tag and cleaned the reported trailing whitespace in api-explorer.html and notifications.html. git diff --check HEAD now reports no trailing-whitespace errors in the HMI changes.

ohwenyi:
Hi, I’ve addressed the requested review changes and pushed the updates to this PR. The API status badge overlap, duplicate <body> tag, and reported trailing whitespace issues have been fixed. The PR is ready for re-review.

src/production/hmi/ui/public/admin/dashboard.html
```diff
@@ -304,11 +306,15 @@ <h6 class="fw-semibold mb-0 fs-4">29/03/2023</h6>
   </div>
 
   <script type="text/javascript">
-    $(document).ready(function() {
-      $("#sidebar").load("/admin/component/sidebar-component.html", function() {
-        $.getScript('/admin/js/sidebarmenu.js');
+    $(document).ready(function () {
```
Vamshi-Gollapelly: Every other page in this PR got moved over to admin-layout.js — admin-nodes, api-explorer, notifications, all three sensor_health pages, the lot — but dashboard.html still has its own version of the same logic sitting here, just rearranged a bit (nesting the header load inside the sidebar callback instead). As far as I can tell that's doing the same job Promise.all already handles in admin-layout.js — wait for sidebar and header both, then kick off app.min.js and sidebarmenu.js — so I'm not seeing a reason this page needs to be the odd one out.

Since the whole point of this PR is consistency across admin pages, could dashboard.html just use the shared file like everywhere else? And if there's actually a reason it can't — some timing thing specific to this page — a quick comment here explaining that would be great, just so it's on record why this one's different.

Vamshi-Gollapelly (dismissed):

src/production/hmi/ui/public/admin/dashboard.html
ohwenyi: Addressed. I updated dashboard.html to use the shared admin-layout.js approach like the other admin pages and removed the duplicated local sidebar/header loading logic. I retested the Dashboard locally and confirmed that the shared layout, navigation and Dashboard functionality still work correctly.

AnDo081105 (commented):
`projects.js` still loads the sidebar, header, and footer independently on lines 8–10. Because this page now also uses `admin-layout.js`, both ready handlers issue concurrent component requests; the later response can replace elements after `app.min.js` and `sidebarmenu.js` attach their handlers, leaving the navigation unresponsive. 

Just remove 
```
// ================================================================
// Layout Includes
// ================================================================
$("#sidebar").load("/admin/component/sidebar-component.html");
$("#header").load("/admin/component/header-component.html");
$("#footer").load("/admin/component/footer-component.html");
```

from `projects.js`

ohwenyi:
Addressed in the latest commit Remove duplicate project layout loading. I removed the duplicate sidebar, header and footer loading calls from projects.js, so admin-layout.js is now the single shared layout initializer for the Project Management page. I retested the page locally and confirmed that the navigation and project functionality still work correctly.

AnDo081105 (approved):
Nice stuff:D

PR #979, opened 2026-08-09
EB/MT: Payload Schema Validation Implementation (C2.2)

Mt-Dkin:
- Added `tests/test_schema_validation.py` to ensure that malformed engine payloads (missing species, confidence out of bounds, invalid types) are correctly rejected with 422 Unprocessable Entity. This prevents corrupt ML model data from polluting the MongoDB datastore.

- Executed inside Docker container: `docker-compose run --rm echo_api sh -c "pip install pytest 'httpx<0.28' && python -m pytest tests/test_schema_validation.py"`

- Output:
 ============================= test session starts ==============================
  platform linux -- Python 3.9.25, pytest-8.4.2, pluggy-1.6.0
  rootdir: /app
  plugins: anyio-4.12.1
  collected 4 items
  tests/test_schema_validation.py ....                                     [100%]
  ======================== 4 passed, 2 warnings in 0.94s =========================

### DIFF
--- a/src/production/backend/tests/test_schema_validation.py
+++ b/src/production/backend/tests/test_schema_validation.py
@@ -1,0 +1,77 @@
+from fastapi.testclient import TestClient
+from app.main import app
+import datetime
+
+client = TestClient(app)
+
+# Note: Using Australian/British spelling for tests where applicable
+def test_valid_payload_normalisation():
+    """Verify that a valid engine payload is processed successfully."""
+    payload = {
+        "timestamp": datetime.datetime.utcnow().isoformat() + "Z",
+        "sensorId": "S123",
+        "species": "Koala",
+        "microphoneLLA": [-33.0, 150.0, 20.0],
+        "animalEstLLA": [-33.01, 150.01, 20.0],
+        "animalTrueLLA": [-33.02, 150.02, 20.0],
+        "animalLLAUncertainty": 5,
+        "audioClip": "base64encodedaudio==",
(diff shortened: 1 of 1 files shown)


src/production/backend/tests/test_schema_validation.py
```diff
@@ -0,0 +1,76 @@
...
+        "animalLLAUncertainty": 5,
+        "audioClip": "base64encodedaudio==",
+        "confidence": 95.5,
+        "sampleRate": 48000
+    }
+    
+    # We expect 422 if the schema is invalid, so if it returns something else 
+    # (like 500 because the DB is down, or 201 Created), it means it passed validation.
+    response = client.post("/engine/event", json=payload)
+    assert response.status_code != 422
```
KCP17, line 25: Nice coverage on the negative cases, and thanks for including the actual pytest output in the description. However, this assertion (!= 422) is looser than it needs to be (your own comment above it points out why, i.e., a 500 from a DB outage, or any other non-422 code, would also pass). Since this test is meant to confirm a valid payload is accepted, I'd suggest asserting the actual expected success status instead, e.g. == 201 (whatever /engine/event actually returns on success) so that it fails loudly if something breaks downstream, instead of silently passing. Everything else looks solid. I'll merge this for now since it's not a bug, but if you have time, please consider amending this line based on my above feedback.

KCP17 (approved):
Approving this for now, but please consider my feedback on line 25, thanks!

PR #985, opened 2026-08-09
Add Sprint 1 species coverage analysis

maniishaww:
Analysed the full merged Project Echo dataset (15,552 recordings across 122 species after fixing a mislabeled species). Found significant class imbalance (up to 300x between most and least represented species), compared coverage across source buckets, identified and fixed a data quality issue (BirdCLEF species using short codes instead of scientific names, causing one species to be undercounted), and included recommendations for balancing the dataset before training (oversampling, augmentation, capping, weighted loss).


KishanFW (approved):
Everything is good

PR #986, opened 2026-08-09
EH/RU/FR-D3: Configure frontend environment variables for HMI

RuhanseeNethusa:
## Summary
- Replaced hardcoded frontend URLs with environment variables.
- Added CLIENT_URL to the example environment configuration.
- Improved deployment configuration for different environments.

## Testing
- Verified the HMI application runs using the configured environment variables.


KCP17 (changes requested):
Two things before this can merge. First, this PR has a real merge conflict with main right now, which needs resolving. Second, the new default (localhost:3001) doesn't match what it's replacing (localhost:9000), or the port I've seen this app documented as using elsewhere (3000). Can you confirm which is actually correct? If it's wrong, donors get redirected to a dead URL after paying. Also, there's no fallback now if CLIENT_URL is unset at runtime (the old hardcoded URL was an implicit safety net), and Stripe will likely reject the session outright if cancel_url comes through as undefined.

Vamshi-Gollapelly (changes requested):
Agree with what KCP17's flagged here, and one thing I can actually settle for you: the port question. I was in another PR's docs earlier tonight (docs/team-guides/Detection_Review_User_Guide.md) and it says outright the HMI is reached at localhost:3000. This PR's new default is localhost:3001. So it's not just "does this look right" there's a documented port elsewhere in the repo that this doesn't match. Worth nailing down which one's actually correct before these merges, because if it's wrong, every donor gets bounced to a URL that isn't serving anything.

The missing-fallback point is the one I'd weigh heaviest on actually. Right now, if CLIENT_URL isn't set and that's a pretty easy thing to forget when standing up a new environment success_url literally becomes the string "undefined/donation-success?..." and cancel_url becomes undefined. I'd expect Stripe to just reject the session outright at that point, which means the entire donation flow goes down, not just a bad redirect. A process.env.CLIENT_URL || 'http://localhost:3000' (or whatever the right default turns out to be) would at least fail safe instead of fail silent.

Also still showing a merge conflict with main, so that'll need resolving regardless.

PR #987, opened 2026-08-09
Add balanced EfficientNetV2 baseline evaluation

mpraveen8535-sketch:
## Summary

- add a reusable EfficientNetV2 TFLite baseline-evaluation workflow
- add the fixed balanced-validation manifest and dataset reference
- include per-file predictions, aggregate metrics, per-species metrics, confusion matrix, and reproducibility evidence
- add focused unit tests for evaluation helpers and validation rules

## Results

The fixed manifest contains 1,142 files across 122 of the 123 configured species.

- top-1 accuracy: 95.18%
- top-3 accuracy: 97.55%
- top-5 accuracy: 97.90%
- macro precision: 95.40%
- macro recall: 94.36%
- macro F1: 94.60%

## Dataset handling

Raw audio and dataset archives are intentionally excluded from GitHub. `DATASET_REFERENCE.md` records the Google Cloud Storage locations, selection rules, manifest hash, shortages, and known limitations.

Because the original training split manifest was unavailable, 737 files use a deterministic reconstructed held-out split and 405 files are marked as unverified backfills that may overlap the original training data. This limitation is preserved in the manifest and reports.

## Validation

- `python -m pytest src/tests/unit/engine/test_selected_model_baseline.py` — 7 passed
- Python compilation check passed
- CLI help smoke test passed
- Git diff whitespace check passed
- payload audit confirmed no raw audio, archives, caches, partial downloads, or local absolute paths


Mangon3 (approved):
Really solid, clean, and well-tested (good pytest assertions).

Some small improvements suggestions from me: 
- **Lead with the held-out number.** `metrics_by_selection_tier.csv` already separates the tiers, and the honest baseline is the reconstructed held-out set: 95.52% top-1 (737 files), worth making that the headline in `RESULTS.md`, with the blended 95.18% as secondary. The 405 unverified backfills sit a bit lower (94.57% top-1) and notably lower on top-3/5 (96.5/96.8% vs 98.1/98.5%), which is consistent with them being harder / possibly training-overlapping (a good reason not to let them dilute the headline).
- **Folder naming:** this lands in `engine/benchmark/…`, right next to the existing `engine/benchmarking/` calibration work, would be great if they are aligned to one home (flag for the Engine cleanup task, not a blocker though).
- **Coordination:** same model as the calibration work but a different manifest, so the accuracy numbers won't line up 1:1. A shared eval manifest would make the two baselines comparable.

Other than that LGTM!

PR #992, opened 2026-08-20
Eb/rs/vulnerability workflow and request size limit

Raghav1103-hub:


Maneesh-Kumar-Maddini (changes requested):
An oversized chunked JSON request returns HTTP 400 rather than the intended 413; this was reproduced during review. Please correct the body-limit handling and cover oversized requests both with and without a content length, including multipart uploads. Also fix the workflow path filter, which references a different filename from the workflow being added. Resolve the merge conflicts and verify the integrated size-limit response contract before approval.

PR #994, opened 2026-08-20
C7.2: Add migration script for timestamp/species index with history tracking

mansipatel765-sk:
Added migration script `001_add_timestamp_species_index.js` to create the compound index `idx_timestamp_species` on the `detections` collection.

Added `migrate.js` runner script that checks `migrations_history` before applying migrations, preventing duplicate runs.

Tested via `docker compose up echo_migrate` — confirmed migration runs successfully and correctly skips re-applying an already-completed migration (verified via `migrations_history` collection).

### DIFF
--- a/src/production/infrastructure/mongodb/migrations/001_add_timestamp_species_index.js
+++ b/src/production/infrastructure/mongodb/migrations/001_add_timestamp_species_index.js
@@ -1,0 +1,11 @@
+// This file creates one index which is a shortcut for searching fast and easily
+// on the detections collection, sorted by timestamp and species. 
+
+const migrationDb = db.getSiblingDB("EchoNet");
+
+migrationDb.detections.createIndex(
+  { timestamp: -1, species: 1 },
+  { name: "idx_timestamp_species" }
+);
+
+print("Done: index created on detections collection");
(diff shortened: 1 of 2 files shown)


Anand152002 (changes requested):
I reviewed you work just some minor changes:

Issue 1 — Migration runner is tightly coupled to /migrations

Please verify that /migrations/001_add_timestamp_species_index.js matches the migration volume/mount configuration in Docker Compose. Since the runner relies on an absolute path, it would be good to confirm this works consistently from the intended migration container.

Issue 2 — No error handling around migration execution

Currently:

load("/migrations/001_add_timestamp_species_index.js");
historyBook.insertOne({ _id: "001", appliedAt: new Date() });

The history entry is inserted only if load() completes, which is good.

However, there is no explicit error handling/logging if the migration fails.

For a migration system, it would be useful to make the failure obvious and prevent the migration runner from appearing successful when the database operation fails.



Else everything looks good thank you

mansipatel765-sk:
Thanks for the review! Fixed both:

Issue 2 (error handling): Added a try/catch around the migration execution in migrate.js. If the migration fails, it now logs a clear `ERROR: Migration 001 failed - <message>` and exits with a non-zero code (`quit(1)`), instead of failing silently. Tested this by temporarily renaming the migration file to force a failure — confirmed the error printed correctly — then restored the file and re-ran it, confirming normal success still works (`Migration 001 done and saved to history`).

Issue 1 (path consistency): Verified via `docker inspect` on the migration container — the local `mongodb/migrations` folder is bind-mounted to `/migrations` inside the container, matching the path used in `load("/migrations/...")`. Confirmed working correctly in the same test above.

Anand152002 (approved):

PR #996, opened 2026-08-27
Fix EchoNet map initialization

KavinduLiyanage04:
Summary

- Fix JavaScript errors preventing the HMI module from loading
- Restore valid module structure
- Allow required browser media resources
- Restore EchoNet map initialization

Testing

- Rebuilt and ran the Docker Compose environment
- Confirmed the EchoNet map displays successfully
- Confirmed the blocking JavaScript errors no longer appear


KishanFW (approved):
Everything is good

PR #998, opened 2026-08-29
Sprint 1 - Dataset Duplicate and Metadata Consistency Validation

deepanshi-k:
## Summary

Completed Sprint 1 Task 3.13 - Dataset Duplicate and Metadata Consistency Validation.

## Work completed

- Implemented duplicate file path detection.
- Implemented potential duplicate metadata record detection.
- Added SHA-256 based audio duplicate detection.
- Added species naming and identifier consistency validation.
- Added checks for missing source, species, species label, and location metadata where available.
- Generated individual CSV reports and a combined structured dataset issues report.

## Current Results

- Total manifest records: 46
- Duplicate file paths: 0
- Potential duplicate metadata records: 10
- Species naming inconsistencies: 41
- Missing metadata records: 15
- Hash/read errors: 46

The hash/read errors occur because the current manifest references `/content/datasets/...` paths and the corresponding physical audio files were not available in my local environment during validation.

## Files Added

- `duplicate_metadata_qa.py`
- `dataset_manifest.csv`
- `reports/dataset_issues_report.csv`
- Individual duplicate, hash, species consistency, and missing metadata reports


Mangon3 (dismissed):
LGTM

AshanVimodh:
Hey @deepanshi-k @Mangon3 - before this merges, I want to flag two issues I found while reviewing that I think need another look. Won't block on the small stuff, but these affect whether the QA results in this PR can actually be trusted.

1. The hash-based duplicate check never ran successfully.
find_duplicate_hashes() hashes df["absolute_path"] as-is, but dataset_manifest.csv stores Colab-specific paths (/content/datasets/...). This is visible in the PR's own output: reports/hash_errors.csv shows all 46 records failing with ERROR:FILE_NOT_FOUND, and reports/duplicate_hashes.csv is empty. This is the one check that looks at actual file content rather than metadata — right now it's a no-op, on this machine or anyone else's.

2. duplicate_records.csv has confirmed false positives. The fingerprint for "duplicate metadata record" is just (species_label, source, size_bytes) - no filename/timestamp. For the sliding-window segmented clips in this dataset, that flags genuinely different audio as duplicates. E.g. under Ceyx azureus: region_19.700-21.700.mp3, region_15.700-17.700.mp3, and region_5.050-7.050.mp3 are three different, non-overlapping 2-second segments - flagged as duplicates purely because they share size_bytes. Same pattern shows up for Capra Hircus, Lalage leucomela, and Uperoleia altissima in the same file. If anyone acts on this report to delete "duplicates," we'd lose real training data.

Given #1, the checked-in dataset_issues_report.csv is also skewed - a large share of the "issues" in it are hash_read_error rows that are an artifact of the path bug, not real dataset problems.

@Mangon3 - since you already approved, would appreciate you taking another pass given the above; happy to discuss if I'm missing context on how the manifest/paths are meant to work here.

Mangon3:
Thanks @AshanVimodh, I checked both branches and you're right.

@deepanshi-k can you perform these 2 fixes:
- Resolve `absolute_path` against a configurable dataset root instead of the baked-in `/content/datasets/` Colab prefix, so the hash check runs anywhere.
- Add `file_name` (or the time window) to the duplicate fingerprint so segmented clips aren't collapsed on shared byte size.

I'll re-approve once these are applied. Thanks for your time.

Mangon3 (changes requested):
Changes stated above

Mangon3 (approved):
LGTM. Thanks for the rerun

PR #999, opened 2026-08-30
D1.3: Add Redis caching for detection retrieval

sharmatushar1912-hub:
## Summary

This PR implements Redis as a caching layer for detection retrieval used by the HMI.

The implementation improves repeated retrieval performance while preventing stale detection data from being served after create, update, or delete operations.

## Changes

- Added Redis client and detection cache helper.
- Added Redis configuration to the backend API container.
- Added `redis==5.0.8` to backend requirements.
- Added query-specific cache keys using filters and pagination values.
- Added a 60-second TTL for detection list cache entries.
- Added cache HIT and MISS handling for detection retrieval.
- Added cache invalidation after successful detection create, update, and delete operations.
- Added JSON-safe serialization for MongoDB ObjectIds.
- Fixed Pydantic configuration for detection response models so ObjectIds are correctly serialized.

## Cache behaviour

For `GET /detections`:

1. The API generates a cache key based on request filters and pagination.
2. Redis is checked first.
3. On a cache HIT, the cached response is returned.
4. On a cache MISS, data is retrieved from MongoDB and stored in Redis.
5. Cache entries expire automatically after 60 seconds.

When detection data changes through create, update, or delete operations, all cached detection list entries are invalidated so stale data is not returned to the HMI.

## Testing

Manual testing completed successfully:

- Redis health check returned `PONG`.
- First detection retrieval produced a cache MISS.
- Repeated retrieval produced a cache HIT.
- Redis detection cache key was created successfully.
- TTL was confirmed at approximately 60 seconds.
- Detection creation invalidated the existing Redis cache.
- Logs confirmed:
  `Detection cache invalidated: 1 key(s)`
- Redis scan confirmed the cached detection key was removed after invalidation.
- The following GET produced a fresh cache MISS.
- Repeated GET then produced a cache HIT.
- Test detection was deleted successfully after validation.
- Final detection retrieval returned `200 OK`.

## Example cache flow

GET /detections
→ Cache MISS
→ MongoDB query
→ Store result in Redis

GET /detections
→ Cache HIT

POST/PATCH/DELETE detection
→ MongoDB changes
→ Detection cache invalidated

Next GET /detections
→ Cache MISS
→ Fresh MongoDB data
→ Cache recreated


AshanVimodh (changes requested):
Hey @sharmatushar1912-hub - before this merges, two things need fixing:

1. Redis errors aren't handled anywhere. get_cached_json/set_cached_json/invalidate_detection_cache in cache.py call Redis directly with no try/except. Since create_detection, delete_detection, and update_detection all call invalidate_detection_cache() after the MongoDB write already succeeds, if Redis is ever down: GET /detections fails outright instead of falling back to Mongo, and a POST/DELETE/PATCH can succeed in the database but still return a 500 to the caller. Could we wrap the Redis calls so a cache failure degrades to "skip the cache" instead of breaking the request?

2. This breaks the existing test suite. tests/test_detection_retrieval.py wasn't updated for caching — none of the list_detections tests mock redis_client, so they'll try to connect to echo-redis:6379 for real and fail in any environment without that host (CI included). Separately, list_detections now returns plain dicts instead of Detection objects, so even with Redis mocked, assertions like result["items"][0].species would raise AttributeError. Four of the five list_detections tests are broken as written.

Once those are sorted this should be good to go.

PR #1002, opened 2026-08-31
EHMI/BG/FR-C2-Device-Detail-Workflow

BalasaitejaG:
Selecting a sensor in the Sensor Health table now opens a detail page for that device, showing current health, hardware and components, location on a map, last audio metadata, and recent audio and reboot history. Reboot and settings actions are available per device, and an unknown sensor ID shows a clear empty state.

Fixes found while building this:

- The sensor table never rendered. script.js declared a top-level const pageState, which every admin page's inline script also declares, throwing a redeclaration error that aborts the whole file.
- Health metrics were all empty. The backend read component readings from "sensorData", but the documents store them under "customProperties".
- Missing values rendered as real data, since Number(null) is 0. Empty readings showed as "0%" battery and "Just now" on Offline devices.
- Every device reported Offline and raised a Critical alert, as status defaulted to Offline when lastSeen was absent and no device sends one. Added a separate Unknown status, excluded from alerts (14 -> 0).
- The location map was blank; its OpenStreetMap iframe is blocked by the HMI CSP. Now uses the OpenLayers build the project already ships.
- The alerts route treated an empty backend response as a failure and substituted demo data, reporting fake alerts once the count hit zero.

Replaced the CPU, RAM, Disk and Uptime columns with Name, Type, Model, Temp and Components; the former had no data source or showed hardware specs as usage percentages. Added regression tests for each fix.


src/production/backend/app/routers/sensors.py
```diff
@@ -314,3 +675,47 @@ def get_recent_reboots(limit: int = Query(50, ge=1, le=200)):
...
+
+
+@router.get("/{sensor_id}", response_description="Get a single sensor's current health, location, and recent audio metadata")
+def get_sensor_detail(
+    sensor_id: str,
+    history_limit: int = Query(10, ge=1, le=50),
+):
+    try:
+        node = Nodes.find_one({"_id": sensor_id})
+        if not node:
```
Vamshi-Gollapelly, line 695: Nice fix for the customProperties/sensorData merge — but none of this (the unit heuristics, GPS fallback chain, or this new /{sensor_id} endpoint) has Python test coverage. Given how much fallback logic is here, could we get a test_sensors.py alongside this?

src/production/hmi/ui/routes/sensor.routes.js
```diff
@@ -0,0 +1,391 @@
...
+      ...settingsUpdate,
+    };
+    return res.json({
+      sensorId,
+      settings: settingsStore[sensorId],
+      source: "demo-fallback",
+    });
+  });
+
+  app.post("/sensors/:sensorId/reboot", async (req, res) => {
```
Vamshi-Gollapelly, line 317: Great regression tests on the JS rendering side (script.load.test.mjs), but this file — the backend-vs-demo-fallback branching — has none. This is exactly the kind of logic that silently broke before (the fake-alerts bug), so a test file here would help catch a regression early.

Vamshi-Gollapelly (commented):
Went through the payment... wait, wrong PR — for this one: "Verified the fixes described (pageState collision, customProperties/sensorData merge, false Offline alerts, the iframe/CSP issue, and the fake-alerts-on-empty bug) against the diff — all check out. Left two comments about missing test coverage on the new backend logic."

src/production/hmi/ui/routes/sensor.routes.js
BalasaitejaG: hi vamshi, I added backend tests for battery status, GPS data, sensor health states, alerts, connected devices, and audio history. I also added HMI route tests to verify live backend data, demo fallbacks, empty responses, and 404 handling. The sensor route helpers were made accessible for testing, and the new tests are included in the regular UI test command. I also cleaned up the existing map test so its expected error does not clutter the test output.

BalasaitejaG: I found some issue with docker, i will be addressing it today and push it back, thanks

src/production/backend/app/routers/sensors.py
```diff
@@ -314,3 +675,47 @@ def get_recent_reboots(limit: int = Query(50, ge=1, le=200)):
...
+
+        payload["lastAudio"] = last_audio_meta
+        payload["recentAudio"] = _recent_audio_events(sensor_id, history_limit)
+        payload["components"] = _normalised_components(node)
+        payload["connectedDevices"] = _connected_devices(node)
+        return jsonable_encoder(payload)
+    except HTTPException:
+        raise
+    except Exception as error:
+        raise HTTPException(status_code=500, detail=f"Error retrieving sensor detail: {str(error)}")
```
AnDo081105: this sends the raw error text to the browser. Please return a short plain message instead and keep the details in the server log. The same pattern repeats in this file around lines 539, 591, 600, 626, 647, 662, and 677.

AnDo081105 (commented):
Hi, I reviewed this PR for coding standards. I found 3 problems.

src/production/hmi/ui/public/admin/sensor_health/script.js
```diff
@@ -91,22 +93,119 @@ function showMessage(elementId, message, type = "success") {
 // API helper
 // ================================================================
 async function apiFetch(path, options = {}) {
```
AnDo081105: this builds a new fetch helper with its own timeout handling. sensor.routes.js and charts.js each build their own copy too. The ticket says reuse the shared client, and FR-D1 says use one common helper. Please use one helper for all three.

src/production/hmi/ui/routes/sensor.routes.js
```diff
@@ -0,0 +1,403 @@
...
+            : `Last contact ${item.lastSeenMinutesAgo} minutes ago`,
+        lastAudioTs: item.lastAudioTs,
+        lastAudioMinutesAgo: item.lastAudioMinutesAgo,
+      });
+    }
+  }
+  return alerts;
+}
+
+async function tryBackend(req) {
```
AnDo081105, line 186: Same problem as my note on script.js line 95: this is a second copy of fetch with timeout. Please use the one shared helper instead of a new one here.

src/production/hmi/ui/public/admin/sensor_health/charts.js
```diff
@@ -1,107 +1,78 @@
...
-    }
+  root.innerHTML = '<p class="card-subtitle">Loading alert summary…</p>';
 
-    const contentType = res.headers.get('content-type') || '';
-    if (!contentType.includes('application/json')) {
-      throw new Error('Alerts chart API did not return JSON.');
-    }
+  try {
+    const controller = new AbortController();
+    const timer = setTimeout(() => controller.abort(), 4000);
```
AnDo081105: Same problem as my note on script.js line 95: this is a third copy of fetch with timeout. Please use the one shared helper instead.

src/production/backend/app/routers/test_sensors.py
```diff
@@ -0,0 +1,318 @@
...
+    def limit(self, _n):
+        return self
+
+    def __iter__(self):
+        return iter(self._docs)
+
+
+class TestAsPctAndFloat(unittest.TestCase):
+    def test_fraction_scales_to_percent(self):
+        self.assertEqual(sensors._as_pct(0.75), 75.0)
```
AnDo081105: these tests check a private helper (_as_pct). Please test what users see instead, like the shown status text and values.

src/production/hmi/ui/routes/sensor.routes.test.mjs
```diff
@@ -0,0 +1,212 @@
...
+  return {
+    status,
+    headers: { get: () => "application/json" },
+    json: async () => data,
+    text: async () => JSON.stringify(data),
+  };
+}
+
+test("alertsFrom skips Online and Unknown, and raises Offline as Critical", () => {
+  const alerts = registerSensorRoutes.alertsFrom([
```
AnDo081105: Same problem: this tests the private alertsFrom helper. Please test the shown alerts instead of the inside helper.

src/production/hmi/ui/public/admin/sensor_health/script.load.test.mjs
```diff
@@ -0,0 +1,155 @@
...
+
+test("script.js runs after a page already declared const pageState", async () => {
+  const source = await readFile(scriptPath, "utf8");
+  const context = vm.createContext(createDomStub());
+
+  // Mirrors the inline block present on every sensor_health admin page.
+  vm.runInContext("const pageState = { resetPageState() {} };", context);
+
+  assert.doesNotThrow(() => {
+    vm.runInContext(source, context, { filename: "script.js" });
```
AnDo081105: Same problem: these tests reach into private page functions through vm. Please test the shown page content instead.

src/production/backend/app/routers/sensors.py
BalasaitejaG: Added test_sensors.py alongside the router. It covers the customProperties/sensorData merge, unit heuristics, GPS fallback chain, status/alert derivation, and the new /{sensor_id} detail endpoint. Run with: python -m unittest app.routers.test_sensors -v from src/production/backend.

src/production/hmi/ui/routes/sensor.routes.js
BalasaitejaG: > Great regression tests on the JS rendering side (script.load.test.mjs), but this file — the backend-vs-demo-fallback branching — has none. This is exactly the kind of logic that silently broke before (the fake-alerts bug), so a test file here would help catch a regression early.

Added routes/sensor.routes.test.mjs for the backend-vs-demo fallback branching, including the empty-alerts regression. It’s wired into npm test.

src/production/backend/backend/project-echo-openapi.json
```diff
@@ -2382,77 +2382,42 @@
         }
       }
     },
-    "/payments/razorpay/orders": {
-      "post": {
+    "/sensors/{sensor_id}": {
```
Vamshi-Gollapelly, line 2385: Merging this as-is will delete the /payments/razorpay/orders, /verify, and /webhook paths that #1003 added to main this branch's copy of the spec doesn't have them. Can you rebase onto latest main and regenerate the OpenAPI file so both feature sets are in the contract?

src/production/backend/app/routers/sensors.py
```diff
@@ -77,15 +117,360 @@ def _extract_battery_pct(node: Dict[str, Any]) -> Optional[float]:
...
+    if last_seen_mins is None:
+        status = "Unknown"
+    elif last_seen_mins <= online_window:
+        status = "Online"
+    elif last_seen_mins <= degraded_window:
+        status = "Degraded"
+    else:
+        status = "Offline"
+
+    if status != "Unknown" and battery_pct is not None and battery_pct <= battery_threshold:
```
Vamshi-Gollapelly, line 444: This overwrites Offline/Degraded with Low Battery too, so a device that's actually offline with a stale low-battery reading shows as Low Battery instead of Offline and the alerts route treats that as a lower severity than Offline. Is that intentional? Might be worth a test either way (there isn't one covering Offline+low-battery together right now).

src/production/backend/app/routers/sensors.py
```diff
@@ -77,15 +117,360 @@ def _extract_battery_pct(node: Dict[str, Any]) -> Optional[float]:
...
+                    candidates.append(number * 100.0)
+                elif 1 < number <= 100:
+                    candidates.append(number)
+            elif number >= 0:
                 candidates.append(number)
 
     if not candidates:
         return None
-    return round(min(candidates), 1)
+    value = min(candidates) if as_percent else candidates[0]
```
Vamshi-Gollapelly, line 130: For non-percentage metrics (e.g. uptime), this just takes the first matching component's value rather than something deterministic if a node ever reports the same key from more than one component, the result depends on array order. Minor, but worth a comment explaining the intent.

Vamshi-Gollapelly (dismissed):

Vamshi-Gollapelly:
This PR and #1027 both rewrite sensor-health.html and sensor_health/script.js independently. This PR adds Type/Model/Temp/Components/Details columns to the table; #1027's row-rendering template only builds 9 columns and doesn't include them (it also adds new KPI cards and status-pill logic this PR doesn't have). Whichever of these merges second will need a real rebase against the other, not just conflict resolution otherwise the Details link or the KPI work will get lost. Might be worth you and Mustafa syncing before both merges.

BalasaitejaG:
> This PR and #1027 both rewrite sensor-health.html and sensor_health/script.js independently. This PR adds Type/Model/Temp/Components/Details columns to the table; #1027's row-rendering template only builds 9 columns and doesn't include them (it also adds new KPI cards and status-pill logic this PR doesn't have). Whichever of these merges second will need a real rebase against the other, not just conflict resolution otherwise the Details link or the KPI work will get lost. Might be worth you and Mustafa syncing before both merges.

Thanks for flagging this. I agree we should avoid merging these independently. Since #1027 includes the KPI/status work, I recommend merging #1027 first. I’ll then rebase #1002 onto the updated base and reconcile `sensor-health.html` and `sensor_health/script.js` so both the KPI/status behaviour and the additional table columns—including the Details link—are retained and tested.

AnDo081105 (approved):

AnDo081105:
You can merge after resolve the conflict

PR #1003, opened 2026-09-01
EH-EB/KD/Fake payment bug

AnDo081105:
# Summary

Adds a Razorpay donation flow that creates authenticated orders through the backend, verifies completed checkouts server-side, and records captures idempotently through either the browser callback or Razorpay webhook.

# Why

- Prevent client-controlled donation records and unverified payment success claims.
- Keep Razorpay credentials and payment verification in the backend.
- Handle cases where a browser callback does not complete.

# What Changed

## Backend payments

- Added `/payments/razorpay/orders`, `/verify`, and `/webhook` FastAPI endpoints.
- Added Razorpay order creation, HMAC signature validation, provider-record verification, and idempotent donation persistence.
- Added payment request/response schemas and MongoDB collections for Razorpay orders and donations.
- Published the updated OpenAPI contract.

## HMI payment flow

- Replaced client-side Razorpay key and direct MongoDB persistence with backend-proxied order creation and verification.
- Stores the signed user token in the session and requires it when creating an order.
- Added rate limiting for order creation and safe upstream error handling.
- Added Razorpay CSP allowances and raw-body webhook forwarding.
- Added environment setup for Razorpay and cookie-secret configuration.

# Functional Fixes

- Enforces configured donation amounts and validates Razorpay IDs, signatures, payment status, currency, and order linkage.
- Prevents duplicate donation records on repeated callbacks or webhook delivery.
- Uses bounded upstream timeouts and returns safe user-facing errors.

# Tests Added

- Backend payment-service coverage.
- FastAPI payment-router coverage.
- HMI Razorpay proxy, rate-limiter, and browser payment-gateway coverage.

# Verification

Automated test files were added, but no test execution result is available from this branch inspection.

# Impact

This is a non-breaking internal payment-flow change. Existing Razorpay checkout now depends on configured backend Razorpay credentials, webhook secret, and HMI session/cookie secret.

# Remaining Follow-Up

- Configure production Razorpay and cookie-secret environment variables.
- Register the HMI webhook endpoint with Razorpay and run the payment test suites in CI.


src/production/backend/app/routers/payments.py
```diff
@@ -0,0 +1,74 @@
...
+    status_code=201,
+    dependencies=[Depends(jwt_bearer)],
+)
+def create_razorpay_order(payload: RazorpayOrderRequest, response: Response):
+    result = payment_service.create_order(payload.amount, _dependencies())
+    return _response(result, response)
+
+
+@router.post(
+    "/verify",
```
Vamshi-Gollapelly, line 49: Noticed /verify (and its HMI proxy) has no auth or rate limit, unlike /orders. The Razorpay signature check covers the security side, but is that gap intentional, or worth adding for consistency?

src/production/hmi/ui/ENV_SETUP.md
```diff
@@ -15,12 +15,18 @@ The `.env` file holds environment-specific settings such as the website address
...
 ```
 
+Generate the Razorpay values in Razorpay Dashboard **Test Mode** for local development. Keep `RAZORPAY_KEY_SECRET` server-side and never commit it to source control.
+
 ## Variable Descriptions
 
 - `API_HOST` — Backend API host. Locally set to `localhost`. Automatically switches to `api-service` in Kubernetes via the ConfigMap.
 - `CLIENT_URL` — Frontend URL used for Stripe redirects and password reset emails. Locally set to `http://localhost:8080`. On the live server, this is provided by the Cloud team.
+- `RAZORPAY_KEY_ID` — Public Razorpay API key ID returned to Standard Checkout. Use a Test Mode key locally and the corresponding Live Mode key in production.
+- `RAZORPAY_KEY_SECRET` — Private Razorpay API key secret used only by the HMI server to create and verify payments. Never expose it to browser code or commit it.
```
Vamshi-Gollapelly: These are only read backend-side (os.getenv in app/routers/payments.py) — the HMI service doesn't use them. Might be worth pointing to the backend's .env.example instead so people don't set them here expecting an effect.

src/production/hmi/ui/public/index.html
```diff
@@ -1165,37 +1163,43 @@ <h2 style="text-align: left; color: rgba(247, 186, 206, 1) !important;">Filter M
...
-            return res.json().then(json => Promise.reject(json));
-          })
-          .then(({ url }) => { window.location = url; })
-          .catch(e => {
-            console.error(e.error);
+        try {
+          const response = await fetch("/api/create-razorpay-order", {
+            method: "POST",
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify({ amount: Number(donorAmount) * 100 }),
```
Vamshi-Gollapelly, line 1180: Does the donation UI only offer the five preset amounts the backend allows ($1/$5/$10/$20/$50), or is there a custom-amount input? If custom amounts are still offered, they'd now get rejected with a 400.

Vamshi-Gollapelly (approved):

src/production/backend/app/routers/payments.py
AnDo081105: for this one, /verify should be open. after someone pays, Razorpay sends a secret signature that proves the payment is real. Requiring the user to still be logged in could cause trouble if their login expires while the payment window is open.

rate limitting is probably good thing to add. Will do soon

src/production/hmi/ui/ENV_SETUP.md
AnDo081105: Yea thats smart. Will also do soon

src/production/hmi/ui/public/index.html
AnDo081105: UI only allow the 5 presets currently. Should probably note this down so that, in the future, if the application decides to change, they would know why

src/production/backend/app/routers/payments.py
Vamshi-Gollapelly: Yeah that makes sense actually — since /orders already ties the order to the logged-in user, /verify doesn't really need its own auth check, the signature is doing that job anyway. And good point about sessions expiring mid-checkout, that would've been an annoying edge case to debug later.

One thing though — since /verify is basically the only endpoint in this flow without auth now, the rate limit becomes more important than just "nice to have," it's the only thing guarding it. Would be good to get that in before this goes live, even as a quick follow-up.

src/production/hmi/ui/public/index.html
Vamshi-Gollapelly: Ah good, that answers it then — if custom amounts aren't exposed anywhere yet the validation isn't blocking anything real, just future-proofing. Appreciate you noting it down, that'll save someone a confusing debugging session later.

nandu1231239 (approved):

PR #1004, opened 2026-09-01
FR-A3: Detection Visualization

Vamshi-Gollapelly:
## What this does

This implements the HMI Planner requirement: the map needs to show live detections with location, confidence, event type, and how old each event is.

Everything lives in `HMI.js` and `index.html`. I didn't pull in any new libraries — the existing OpenLayers map and `hmiState`/layer patterns already had what I needed, so I built on top of those instead of reinventing anything.

## How the acceptance criteria are covered

- [x] **Vocalization markers now show up at the right spot.** Turns out `_addVocalizationFeature` was plotting these at the simulator's ground-truth coordinates (`locationLat/locationLon`), not the estimated location (`estLat/estLon` from `animalEstLLA`). That's fixed now — it only falls back to the true location if the estimated fields aren't there, which does happen with the page's hardcoded sample data.
- [x] **Truth and vocalization events look different on purpose.** Vocalizations get an amber circle, truth/movement events get a teal diamond, both sitting under the existing species icon. I went with shape *and* color rather than just color — didn't want this to fall apart for anyone with color vision differences.
- [x] **Confidence/uncertainty gets a visual ring.** Each vocalization marker has a translucent ring around it — an actual `ol.geom.Circle` in map metres, so it scales properly as you zoom instead of staying a fixed pixel size. It's sized off the real `animalLLAUncertainty` value. Truth events don't get a ring since they're ground truth — no uncertainty to show.
- [x] **Old events fade out.** I kept the existing `hmiState.liveWindow` (still 60s by default, documented right where it's defined in `index.html`) rather than adding a second config value — didn't see a reason to duplicate it. Markers stay fully visible until they're halfway through that window, then fade down to a floor right as the existing purge logic removes them.

## One thing worth flagging: the confidence units

I got this wrong on my first pass, and it's worth mentioning in case anyone else assumes the same thing. I initially treated `locationConfidence` (`100 - animalLLAUncertainty`) as a 0–100 score. It's not — I traced it back through the edge inference README ("GPS accuracy estimate in metres") and the engine/simulator code, and `animalLLAUncertainty` is an actual distance in metres. So the ring radius now uses that metres value directly (capped at 300m), rather than the percentage math I started with. Left `locationConfidence` alone since the existing popup still uses it for its "Location Confidence %" label.

## On performance

Switched from static per-feature `Style` objects to style *functions*, so OpenLayers recomputes the fade opacity itself during its normal repaint cycle for whatever's on screen — no need to manually rebuild the whole map every tick. I also extended the per-tick `layer.changed()` call (it was only hitting truth layers before) to cover vocalization layers too, so the fade keeps pace with the 1-second poll.

## Assumptions I'm making (flag if wrong)

- "Estimated location" = `animalEstLLA`. True location (`animalTrueLLA`) only sticks around on the feature as fallback/debug info, never for placing the marker.
- "Movement/truth" = `/movement_time` events, "vocalization" = `/events_time` — this matches how the code already names things (`movementEvents`/`vocalizationEvents`, the `_truth` layer suffix).

## What I actually tested

Ran this locally with `npm run dev` (didn't spin up the full Docker/EchoNet stack). Confirmed visually: shapes/colors read correctly, rings stay legible when they overlap, clicking a marker still opens the right detail panel with real data, and the fade behaves correctly when I manually bump `hmiState.currentTime` forward in the console.

Haven't tested this against real live acoustic detections yet — that needs the full stack running per `src/production/README.md`, and the sample dataset doesn't really vary enough (ring sizes, event timing) to stress-test it properly.

## Not related to this PR, but noticed it

`index.html` references a handful of CSS files — `HMI.css`, `HMI-menu.css`, `HMI-controls.css`, `HMI-audio.css`, `HMI-weather.css` — that don't actually exist in `public/css/` (only `HMI_style.css` and `nodes.css` are there). Didn't touch it since this feature's styling is all inline OpenLayers, but flagging it in case it's worth its own ticket.


BalasaitejaG:
Looks good overall. One small thing: if all three Docker build attempts fail, the workflow may still continue instead of failing. Could we make it exit with an error after the last failed attempt?

Vamshi-Gollapelly:
> Looks good overall. One small thing: if all three Docker build attempts fail, the workflow may still continue instead of failing. Could we make it exit with an error after the last failed attempt?

fixed it now. added an exit 1 for the last attempt so if all 3 builds fail the step actually fails instead of just sleeping and reporting green. had to fix an unrelated indentation bug in the same spot too. CI's passing now

src/production/hmi/ui/public/js/HMI.js
AnDo081105, line 950: In here, 929 to 950 dereferences missing/null LLA arrays before fallback runs

src/production/hmi/ui/public/js/HMI.js
```diff
@@ -1216,24 +1358,39 @@ function _addTruthFeature(hmiState, entry) {
...
+  // It is kept on the feature (animalTrueLon/animalTrueLat) for debugging
+  // only and is never used to place the marker or the confidence ring.
+  // If the estimated fields are ever missing, fall back to the true location
+  // rather than silently dropping the marker.
+  const estLon = Number.isFinite(entry.estLon) ? entry.estLon : entry.locationLon;
+  const estLat = Number.isFinite(entry.estLat) ? entry.estLat : entry.locationLat;
+
   const feature = new ol.Feature({
-    geometry:          new ol.geom.Point(ol.proj.fromLonLat([entry.locationLon, entry.locationLat])),
+    geometry:          new ol.geom.Point(ol.proj.fromLonLat([estLon, estLat])),
```
AnDo081105, line 1374: here, it validates lat and lon independently, allowing an estimated longitude to be combined with a true latitude.

AnDo081105 (commented):
Validate complete coordinate pairs, including ranges, then select estimated, fallback, or skip the malformed event.

AnDo081105 (dismissed):

src/production/hmi/ui/public/js/HMI.js
Vamshi-Gollapelly: Fixed both the vocalization converter was crashing on null LLA arrays, so it reads them safely now. And for the marker placement, it was mixing estimated/true coords independently, so now it treats lat+lon as one pair, uses the estimate if valid, falls back to true location if not, and skips the event entirely if neither is usable.

KishanFW (approved):
All good

PR #1008, opened 2026-09-04
FR-D2: Back the admin notifications with real data and render them sa…

Alexander-Petkovski:
FR-D2: Notification Reliability and Safety

## Summary

The admin notification list was a hardcoded array of three dummy entries sitting in `notifications.js`. Marking something read, deleting it, or marking all read only mutated that array, so every action came back undone on the next refresh. The message text was also interpolated into an HTML template string, and the unread badge had nothing valid to render into.

## Acceptance criteria

| Criterion | Where |
|---|---|
| Notifications loaded from or synchronised with a backend source | `/api/notifications` in `server.js`, feed derived from real donation and user records |
| Read, unread, delete and mark-all-read persist after refresh | Flags stored in a `notificationState` collection, verified through a container restart |
| Messages rendered using safe DOM text handling | `notifications.js` rebuilt on `createElement` / `textContent`, no HTML strings |
| Unread badge stays synchronised with the list | Both drawn in one render pass from the same response, and the badge element fixed |

## Backend

New routes in `server.js`:

- `GET /api/notifications`
- `PATCH /api/notifications/read-all`
- `PATCH /api/notifications/:id/read`
- `DELETE /api/notifications/read`
- `DELETE /api/notifications/:id`

Every action returns the refreshed list, so the browser never keeps a second copy of the state to keep in step by hand.

The feed is built from records we already hold, donations and user registrations, rather than invented data. Ids are derived from the source document (`donation:<id>`, `user:<id>`) rather than generated, otherwise the read state would not line up with the same notification on the next request and everything would look unread again after a reload.

Read and deleted flags live in their own `notificationState` collection, which keeps us from writing UI flags onto the donation and account documents.

Feed shaping and the state join were kept out of `server.js` and put in `services/notifications.js`, so they can be tested without Mongo or a running server.

`DELETE /api/notifications/read` is declared before `DELETE /api/notifications/:id`, otherwise "read" gets swallowed as an id.

## Safe rendering

`notifications.js` no longer builds any element from an HTML string. Notification text is drawn from donation and account records, which makes it effectively user supplied, so a crafted email address dropped into `innerHTML` would have run script in the admin console. Everything is created with the DOM API and written in with `textContent`.

The icon is now chosen from the notification type against a fixed list in the file rather than taken from the response, so nothing outside `notifications.js` decides what class lands on an element.

## Unread badge

The badge was `<class id="notification-badge">`. There is no such element, so there was nothing for the count to render into. It is now a `span`, styled alongside the header component so it looks the same on every admin page that includes the header.

The list and the badge are drawn in a single render pass from the same server response, so they cannot disagree. The previous code updated them from four separate places and leaned on a 100ms `setTimeout` to catch the header finishing loading.

The header is injected after this script runs, so the badge is painted when a `MutationObserver` sees it arrive. The observer disconnects at that point on purpose: the badge sits inside `#header`, so leaving it connected means the badge write counts as a mutation and re-triggers the render in a loop.

## Testing

`npm test` is 32 passing, 17 of them new in `tests/notifications.test.mjs`, covering feed shaping, id derivation, the read and deleted join, sorting, and that the unread count is derived from the same list it is counting.

Verified against the live docker-compose stack:

- 12 notifications built from the real donation and user records
- mark read, delete, mark all read and delete read each persisted across separate requests **and a full container restart**, so the state is genuinely in Mongo
- through the UI: marking one read took the badge from 12 to 11 and matched the list exactly, and survived a page refresh
- mark all read left the badge empty and hidden, and delete read showed the empty state

Safe rendering was checked with a real payload rather than a shimmed DOM. Passing `<script>...</script>` as a notification message through the real `buildNotificationItem` and attaching the result to the page produced:

- no script execution
- zero `img` and zero `script` elements created
- `textContent` equal to the raw payload, and `innerHTML` beginning `&lt;img`
- the message as a text node

Any notification state created while testing was cleared afterwards.

## Worth a look

`/api/notifications` is not behind `checkUserSession`, matching `/donations` and `/cumulativeDonations` which are already open. I did not add the existing middleware to it because it answers with a redirect to `/login`, which is wrong for an API: axios follows the redirect and the caller ends up parsing the login page as JSON, which is the same failure that was fixed in the sign-in route in #946.

Flagging it rather than fixing it here because authorisation is not in the acceptance criteria for this ticket, and because the session check reads a single shared `JWT` key from Redis rather than a per-user session, so a guard on this endpoint would not mean much until that changes. Happy to pick it up as its own piece of work if you would rather it were handled now.

### DIFF
--- a/src/production/hmi/ui/middleware/index.js
+++ b/src/production/hmi/ui/middleware/index.js
@@ -79,6 +79,57 @@ function _isPublicRoute(path) {
 const checkUserSession = createCheckUserSession(client, _isPublicRoute);
 
 // ─────────────────────────────────────────────────────────────────────────────
+// API session guard
+// ─────────────────────────────────────────────────────────────────────────────
+
+/**
+ * Build a session guard for routes that answer with JSON.
+ *
+ * checkUserSession above redirects to /login when there is no session, which is
+ * right for a page but wrong for an API: axios follows the redirect, gets the
+ * login page back with a 200, and the caller ends up parsing HTML as JSON. That
+ * is the same failure that had to be fixed in the sign-in route. This replies
+ * with a status the caller can actually act on instead.
+ *
+ * The token lookup is passed in so the guard can be tested without Redis.
+ *
+ * @param {() => Promise<string|null>} getToken
+ * @returns {import('express').RequestHandler}
+ */
+function createApiSessionGuard(getToken) {
+  return async function apiSessionGuard(req, res, next) {
+    try {
+      const token = await getToken();
+
+      if (!token) {
+        return res.status(401).json({
+          error: "Your session has expired. Please sign in again."
+        });
+      }
+
+      return next();
+    } catch (error) {
+      // Redis being unreachable is our problem rather than the caller's, so this
+      // is a 503 and not a 401. A 401 would tell them to sign in again, which
+      // would not help and would lose whatever they were doing.
+      console.error("API session check failed (Redis error):", error);
+      return res.status(503).json({
+        error: "Unable to verify your session. Please try again shortly."
+      });
+    }
+  };
+}
+
+/**
+ * Session guard for JSON API routes, backed by the same Redis JWT that
+ * checkUserSession reads.
+ */
+const requireApiSession = createApiSessionGuard(async () => {
+  await ensureRedisConnected();
+  return client.get("JWT");
+});
+
+// ─────────────────────────────────────────────────────────────────────────────
 // Session helpers
 // ─────────────────────────────────────────────────────────────────────────────
 
@@ -110,6 +161,8 @@ async function clearUserSession() {
 module.exports = {
   verifySignUp,
   checkUserSession,
+  createApiSessionGuard,
+  requireApiSession,
   clearUserSession,
   client,
 };
--- a/src/production/hmi/ui/package.json
+++ b/src/production/hmi/ui/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "server.js",
   "scripts": {
-     "test": "node --test tests/nodes-overlay.test.mjs public/js/audio_recorder.test.mjs tests/razorpay-payment.test.js tests/paymentgateway.test.js",
+     "test": "node --test tests/nodes-overlay.test.mjs tests/notifications.test.mjs public/js/audio_recorder.test.mjs tests/razorpay-payment.test.js tests/paymentgateway.test.js",
     "dev": "nodemon --inspect=0.0.0.0 server.js"
   },
   "author": "",
--- a/src/production/hmi/ui/package.json
+++ b/src/production/hmi/ui/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "server.js",
   "scripts": {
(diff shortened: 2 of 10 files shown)


AnDo081105 (changes requested):
Requesting changes for one point: the new notifications API is unauthenticated and returns PII.

`GET /api/notifications` (plus the PATCH/DELETE routes) in `src/production/hmi/ui/server.js` has no session check, and the feed now returns real donor emails and registered user emails/usernames. Before this PR the list was 3 hardcoded dummy entries, so this newly exposes real PII to anyone who can reach the endpoint.

I saw the note in the PR description about deferring auth (redirect-based middleware from #946, shared JWT concern). That context is fair, but since this PR expands exposure vs the old dummy list, I need either:

1. a proper guard on these routes that returns 401 JSON for unauthenticated API callers (not a login-page redirect), or
2. explicit maintainer sign-off recorded on the ticket that leaving it open is accepted risk, with a follow-up issue for auth.

Happy to approve once one of those is in place. Everything else in the PR looks good to me.

AnDo081105:
Also resolve the merge conflict. rebase to main then run test before merge

Thank youu

Alexander-Petkovski:
Guard has been been put in, going with option 1 "a proper guard on these routes that returns 401 JSON for unauthenticated API callers (not a login-page redirect)"

All five routes now require a session. I did not reuse `checkUserSession` because it redirects to `/login`, which an API caller cannot act on: axios follows the redirect, gets the login page back with a 200, and tries to parse HTML as JSON. That is the same failure that had to be fixed in the sign-in route.

- `createApiSessionGuard` answers `401` with a JSON body and never redirects. The token lookup is injected so it can be tested without Redis.
- `requireApiSession` wires it to the same Redis JWT the page guard reads.
- A Redis outage returns `503` rather than `401`. Telling someone to sign in again would not help and would lose whatever they were doing.
- The browser now shows the reason the API sends back, so an expired session reads as such.

Verified against the live stack through docker:

- no session: `401`, `application/json`, zero redirects, on all five routes
- the 401 body carries no donor or account data
- with a session: the feed and all four actions still work and persist
- the page with no session shows "Your session has expired. Please sign in again." and renders no PII

Six new tests in `tests/api-session-guard.test.mjs` cover the pass through, the 401 and its JSON body, that it never redirects, the 503 on a lookup failure, and that the underlying error is not leaked. 51 pass in total.

Rebased onto main. One conflict in `package.json`, where the razorpay tests and the notification tests had both been added to the `test` script. Kept both.

Vamshi-Gollapelly (approved):
The session guard addresses the PII exposure concern raised earlier createApiSessionGuard returns 401 JSON (no redirect) on all five notification routes, backed by the same Redis JWT the page guard uses, with a 503 fallback on Redis outages rather than a misleading 401. Verified against the live stack: unauthenticated calls get zero PII in the 401 body, and the 6 new tests in tests/api-session-guard.test.mjs cover the pass-through, 401/JSON shape, no-redirect behavior, 503 on lookup failure, and no error leakage.

Rebase onto main looks clean (package.json conflict resolved by keeping both test script additions), and all 51 tests pass.

Nice work tightening this up, LGTM.

AnDo081105 (approved):

PR #1010, opened 2026-09-04
EB/AVD/Add typed Backend Settings model, migrate scattered config/secrets

AshanVimodh:
## Summary
- Add `app/config.py`, a single `pydantic.BaseSettings` model covering Mongo, Redis, MQTT, API, timeouts, auth, and Twilio/mail settings — replacing scattered `os.getenv()` calls and hardcoded values across `database.py`, `insights.py`, `hmi.py`, `middleware/auth.py`, `two_factor.py`, `utils/sms.py`, `routers/sim.py`, `logging_config.py`, and `main.py`.
- Migrates real hardcoded secrets (JWT signing secret, Twilio SID/token/phone, Gmail SMTP password) out of source into the settings model. Same literal values, just relocated — no rotation, since they're already in git history.
- Required fields (`mongodb_uri`, `user_mongodb_uri`, `jwt_secret`) fail fast with a clear `pydantic.ValidationError` at startup if missing, instead of surfacing later as an unrelated crash.
- Fixes two latent bugs found during consolidation: `insights.py` was opening its own Mongo connection against a separate `MONGO_URI` env var pointing at a non-existent `mongo` host; `hmi.py` called itself via two different hardcoded URLs. Both now go through shared settings.
- Tightens the `pydantic` pin to `>=1.10,<2.0.0` (no `pydantic-settings` package needed — `BaseSettings` ships in `pydantic` itself in v1) and drops the unused `python-decouple` dependency.
- Adds `test_config.py` (7 unit tests: fail-fast on missing required fields, defaults, type coercion).

This sets up the field names later Sprint 2 tasks (C1/C2/C7/C9/C10/C11) already expect from a shared `settings` object (`settings.redis_host`, `settings.mqtt_detection_topic`, `settings.request_timeout_seconds`, `settings.slow_operation_ms`, `settings.cache_ttl_seconds`), and gives C5 (secret-file loading) fields to extend.

## Test plan
- [x] `docker compose up -d --build echo_api` — clean startup, no `ValidationError`
- [x] Temporarily removed `JWT_SECRET` from `.env` — confirmed fail-fast `ValidationError` at import time, then restored
- [x] `docker exec ts-api-cont python -m unittest test_config -v` — all 7 tests pass
- [x] Smoke-tested `/docs`, `/detections`, `/insights/overview`, `/insights/species` — no regressions
- [x] Verified JWT sign/decode round-trip works with the relocated secret
- [x] `.env` changes are local-only (gitignored) — only placeholder `.env.example`/`.env_example` templates are committed, no secrets in the diff


nandu1231239 (approved):

nandu1231239:
please resolve the conflicts

PR #1012, opened 2026-09-05
EH/OW/FR-B3 Event Audio Playback Integration

ohwenyi:
Implements FR-B3 Event Audio Playback Integration.

Changes:

Added user-visible animal audio status messages for empty, loading, ready, playing, and unavailable states.
Added stale-event protection so an older audio request cannot overwrite a newer event selection.
Preserved the existing shared event-audio workflow used by playback and spectrogram rendering.
Added status reset handling when playback stops or finishes.

Validation:

Production HMI runs successfully on the FR-B3 branch.
Frontend states validated using audio_spectrogram_prototype.
Animal workflow tests for shared decoded audio, stale selection handling, and decode/route errors pass.
Full local end-to-end event audio playback is limited because the local sample markers are not connected to real audio files and production audio depends on Google Cloud, as confirmed by the team lead.


src/production/hmi/ui/public/js/HMI.js
```diff
@@ -1408,11 +1431,37 @@ function createMapClickEvent(hmiState) {
...
-          void animalSpectrogramWorkflow?.select(selectedVocalizationEventId);
+          const eventId = selectedVocalizationEventId;
+
+          setAnimalAudioStatus("Loading audio...");
+
+          if (!animalSpectrogramWorkflow) {
+            setAnimalAudioStatus("Audio unavailable for this detection.");
+          } else {
+            void animalSpectrogramWorkflow
+              .select(eventId)
```
Vamshi-Gollapelly, line 1444: Should we add a .catch() here? If select() rejects instead of resolving to a falsy value on failure, the status stays stuck on "Loading audio..." with no path back to "ready" or "unavailable," and it'll throw an unhandled promise rejection. Otherwise this looks good — the stale-event guard on line 1447 is correct.

src/production/hmi/ui/public/js/HMI.js
```diff
@@ -899,6 +908,12 @@ export function stopAudioPlayback(updateAnimation = true) {
     void activeAudioContext.close().catch(() => {});
   }
   activeAudioContext = null;
+
+  if (selectedVocalizationEventId !== null) {
```
Vamshi-Gollapelly, line 912: Same stale-event concern as the async decode path: this only checks selectedVocalizationEventId !== null, not whether it still matches the event this audio belongs to. If a user switches to a different event while the previous one's audio is still playing, this onended firing later will overwrite the new event's status with "Audio ready..." even though it's stale. Same question applies to the equivalent check in stopAudioPlayback. Should these also capture and compare the event id, like the. then() block does?

Vamshi-Gollapelly (dismissed):

src/production/hmi/ui/public/js/HMI.js
```diff
@@ -1408,11 +1431,37 @@ function createMapClickEvent(hmiState) {
...
+                if (decodedAudio) {
+                  setAnimalAudioStatus(
+                    "Audio ready. Press the audio button to play."
+                  );
+                } else {
+                  setAnimalAudioStatus(
+                    "Audio unavailable for this detection."
+                  );
+                }
+              });
```
AnDo081105, line 1458: this `select(eventId).then(...)` has no `.catch(...)`. Right now `select()` never fails, so it works. But if it ever fails, the text will stay stuck on "Loading audio...". Please add a short `.catch()` that shows "Audio unavailable for this detection." Same is needed for `getSelectedAudio().then(...)` near line 964.

src/production/hmi/ui/public/js/HMI.js
```diff
@@ -926,10 +941,17 @@ function playDecodedAudio(decodedAudio) {
...
+    if (selectedVocalizationEventId !== null) {
+      setAnimalAudioStatus(
+        "Audio ready. Press the audio button to play."
+      );
+    }
   };
   activeAudioContext = context;
   activeAudioNode = source;
   source.start();
+  setAnimalAudioStatus("Playing audio...");
```
AnDo081105, line 954: Small fix needed: when the audio button is pressed but there is no audio (around line 964-966), we only stop the animation. The text still says "Audio ready..." even though nothing plays. That can confuse users. Please also set the text to "Audio unavailable for this detection." there, same as the code below does.

src/production/hmi/ui/public/js/HMI.js
```diff
@@ -926,10 +941,17 @@ function playDecodedAudio(decodedAudio) {
     source.disconnect();
     void context.close().catch(() => {});
     muteAudioAnimation();
+
+    if (selectedVocalizationEventId !== null) {
+      setAnimalAudioStatus(
+        "Audio ready. Press the audio button to play."
```
AnDo081105, line 947: Small cleanup: the text "Audio ready. Press the audio button to play." appears 3 times, and this whole block appears twice (here and in `stopAudioPlayback` above). The text "Audio unavailable for this detection." also appears 2 times. If the wording ever changes, we have to change it in many places. Please use shared texts and one small helper instead.

Suggestion:
```js
const ANIMAL_AUDIO_READY = "Audio ready. Press the audio button to play.";
const ANIMAL_AUDIO_UNAVAILABLE = "Audio unavailable for this detection.";

function markAnimalAudioReady() {
  if (selectedVocalizationEventId !== null) {
    setAnimalAudioStatus(ANIMAL_AUDIO_READY);
  }
}
```
Then replace both blocks (in `stopAudioPlayback` and in `source.onended` here) with just `markAnimalAudioReady();`, and replace both "unavailable" spots with `setAnimalAudioStatus(ANIMAL_AUDIO_UNAVAILABLE);`.

src/production/hmi/ui/public/js/HMI.js
ohwenyi: Addressed in the latest commit. I added a .catch() to the select(eventId) flow and kept the stale-event guard, so a rejected request now shows the unavailable-audio state instead of remaining on “Loading audio...”.

src/production/hmi/ui/public/js/HMI.js
ohwenyi: Addressed in the latest commit. Playback now tracks the event ID associated with the active audio. The ended/stopped callbacks only update the ready state when that event still matches the currently selected vocalisation, preventing stale audio from overwriting the newer event state.

src/production/hmi/ui/public/js/HMI.js
ohwenyi: Addressed in the latest commit. I added .catch() handling to both the select(eventId) flow and the getSelectedAudio() playback flow. If either request rejects, the HMI now shows the unavailable-audio state, while stale-event checks prevent an older event from overwriting the current selection.

src/production/hmi/ui/public/js/HMI.js
ohwenyi: Addressed in the latest commit. The playback handler now checks for a missing decoded audio result and sets the status to “Audio unavailable for this detection.” instead of leaving the UI in the ready state.

src/production/hmi/ui/public/js/HMI.js
ohwenyi: Addressed in the latest commit. I added shared ANIMAL_AUDIO_READY and ANIMAL_AUDIO_UNAVAILABLE constants and introduced markAnimalAudioReady(eventId) to centralise the ready-state check. I also replaced the duplicated ready/unavailable status blocks with the shared helper and constants.

KishanFW (approved):

PR #1013, opened 2026-09-05
EB/AVD/Docker Secrets, File-Based Secret Loading

AshanVimodh:
## Summary
- Extends the C4 typed `Settings` model so production-sensitive values (`MONGODB_URI`, `USER_MONGODB_URI`, `JWT_SECRET`, `TWILIO_ACCOUNT_SID`, `TWILIO_AUTH_TOKEN`, `MAIL_PASSWORD`) can be loaded from a mounted secret file instead of a plain environment variable.
- Uses an explicit `<NAME>_FILE` convention (e.g. `JWT_SECRET_FILE=/run/secrets/jwt_secret`) rather than pydantic's implicit `secrets_dir`, wired in via `Config.customise_sources`.
- Precedence: `<NAME>_FILE` (if set) > plain `<NAME>` env var > `.env` file. This lets local dev keep using plain `.env` values while production can mount real secrets as files.
- Documents the convention with placeholder examples in `.env.example`.

## Why
Part of Sprint 2 task **C5 – Docker Secrets / File-Based Secret Loading**, which builds directly on the C4 settings model (#<C4 PR number>) so secrets never need to be committed, hardcoded, or passed as plain env vars in production.

## Testing
- 5 new unit tests in `test_config.py` covering:
  - secret file used when the plain env var is absent
  - secret file takes precedence over a plain env var when both are set
  - fallback to the plain env var when no `_FILE` var is set
  - an unset `<NAME>_FILE` is not mistaken for an empty file path
  - secret file content is whitespace-trimmed
  - secret values never leak into a `ValidationError` message
- All 13 `test_config.py` tests (8 existing C4 + 5 new) pass, both locally and inside the real `ts-api-cont` container (Python 3.9, pinned `pydantic` 1.10.x).
- Verified live against the full Docker Compose stack:
  - `docker compose up -d --build echo_store echo-redis echo_mqtt echo_api` — API started cleanly, no `ValidationError`, no secrets in logs.
  - `/docs` and `/openapi.json` responded correctly.
  - Mounted a placeholder secret file and set `JWT_SECRET_FILE` on a one-off `docker compose run` — confirmed `settings.jwt_secret` read the file content, overriding the plain `JWT_SECRET` env var.


nandu1231239 (approved):

nandu1231239:
please resolve the conflicts

PR #1015, opened 2026-09-05
IoT/nm/esp32code

gorillaPet:


src/production/iot/ESP-32-S3 Code/main/config.h
```diff
@@ -0,0 +1,100 @@
...
+//   HFR   - rms(sample-to-sample difference) / rms(signal). A cheap
+//           spectral centroid: for a tone at f it equals 2*sin(pi*f/fs).
+//           0.45 is roughly a 1.2 kHz centroid. Below that the energy is
+//           all low-frequency (steady wind, rumble, hum) and not bird-shaped.
+//
+// These are starting points from synthetic test signals. Watch the [gate]
+// lines against clips you can actually hear and retune.
+#define GATE_MIN_RMS       150.0
+#define GATE_MAX_CLIP_PCT  1.0
+#define GATE_MIN_HFR       0.00
```
KCP17, line 81: This is set to 0.00, which means the HFR check (st.hfr < GATE_MIN_HFR) can never actually trigger since hfr can't go negative, so the wind/rumble gate is effectively off right now. Is this intentional (waiting on field data before enabling it) or a leftover from testing? Approving regardless since it's not a bug, just something worth confirming.

KCP17 (approved):
Feedback and reason for approval are put on code line 81.

src/production/iot/ESP-32-S3 Code/main/config.h
gorillaPet: Hey, it's mostly intentional, I wasn't planning to do any edge case filtering as part of my technical side but wanted to leave the basis of a rudimentary filter there. So I left the gate off deliberately but it can be adjusted as a later task in T3. Thanks.

PR #1016, opened 2026-09-05
EE/QDN: Fix LMDB cache collision in reproducible_training_pipeline dataset

25qdatttt:
### Required Work

**Own the pipeline architecture and review Oh Fu Sheng's implementation of the deferred items.**

Reviewed the current pipeline end-to-end by running it against real (non-synthetic) data instead of just reading code. This surfaced 3 real bugs, one fixed here:

- **Fixed**: `train_dataset`/`val_dataset` each lazily open their own `lmdb.Environment` at the same cache path. With `num_workers=0` both live in one process, and lmdb refuses to open the same environment twice there; crashed the first real training run as soon as validation started. Fixed by giving each split its own cache subdirectory (`dataset.py`).
- **Found, not fixed here** (documented in `docs/sprint2/SPRINT2_LOG.md`): a race condition with `num_workers>0` (workaround: use `num_workers=0`), and a CPU OOM in `_evaluate()`/`test()` when a file's chunk count is unusually large (no sub-batching for long files).

Also pre-validated the QAT path Oh Fu Sheng is implementing under 3.11, ahead of his fix, so the findings are ready for him:

- `prepare_qat_fx`'s default `input_size=(1, 3, 32, 32)` in `model/quant.py` is wrong: real spectrogram input is `(1, 1, 384, ~200)` (1 channel, not 3). `model/__init__.py` calls it without passing `input_size`.
- After `model.quantise()` (`convert_fx()`), the model can't run inference on a plain float tensor: no `torch.quantize_per_tensor()`/QuantStub step before input reaches the quantised model (`aten::qscheme` not implemented for CPU backend).

**Run the first full (non-synthetic) training on the current dataset** (balanced set from Manisha/Raveesha not yet available, per the plan's fallback clause).

Ran a real 15-epoch training on all 127 species (`efficientnet_v2`, GPU): `val_acc` 6.0% to 10.71%, `val_loss` best 4.1150 at epoch 10. Checkpoint + TensorBoard log produced. Full run details, per-epoch table, and timings in `docs/sprint2/SPRINT2_LOG.md`.

### Expected Sprint 2 Outputs

- **Validated QAT path**: validated, **not working end-to-end** yet (see findings above); precise root cause identified and handed to Oh Fu Sheng.
- **First full training log**: done, 15-epoch real run, see log.
- **Reviewed, working pipeline**: done, pipeline verified to run end-to-end on real data + GPU after the LMDB fix; 2 additional bugs documented as known follow-ups.

Full technical log (bug repro steps, decisions, rationale) in `docs/sprint2/SPRINT2_LOG.md`.


Mangon3 (approved):
LGTM

PR #1017, opened 2026-09-05
Fix CI: bump Engine.test.Dockerfile off EOL Debian bullseye to bookworm

AshanVimodh:
Engine.test.Dockerfile was pinned to python:3.10-slim-bullseye (Debian 11), which reached full end-of-life on 2026-08-31. Its apt archive is being frozen/migrated, causing apt-get install to intermittently 404 on packages that are still listed in a freshly-updated index - specifically gnupg2/gpgsm, pulled in as a dependency of plain gnupg, which broke the combined base-libs + Google Cloud CLI install step in CI's "Docker Image CI" workflow.

Bump the base image to python:3.10-slim-bookworm (Debian 12, currently supported), and switch the Google Cloud CLI install from apt/apt-key to Google's own tarball, removing the gnupg/apt-key dependency chain that was actually failing rather than working around it.

Also renames libopenexr25 -> libopenexr-3-1-30, since bookworm ships OpenEXR 3.x under a different package name.

Verified locally with the same flags CI uses (--no-cache --force-rm, fresh pull from live mirrors): image builds clean, `gcloud --version` works, and tensorflow/opencv/pymongo/paho-mqtt/librosa all still import correctly inside the container after the base image bump.

### DIFF
--- a/src/production/engine/Engine.test.Dockerfile
+++ b/src/production/engine/Engine.test.Dockerfile
@@ -1,5 +1,8 @@
 # Builder (Compilers and heavy lifting)
-ARG BASE_IMAGE=python:3.10-slim-bullseye
+# bullseye (Debian 11) reached full end-of-life 2026-08-31; its apt archive
+# is being frozen/migrated, which breaks package installs unpredictably.
+# bookworm (Debian 12) is the current supported release.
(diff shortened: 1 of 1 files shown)


Mangon3 (approved):
Thanks for the fix!

PR #1018, opened 2026-09-06
EB/JB/B5.1 API Response Compression

jayden-s222587117:


Maneesh-Kumar-Maddini (changes requested):
On this branch, response compression interacts incorrectly with error middleware. A large JSON error was reproduced as an empty response with a gzip header because the middleware consumed the compressed body and returned the exhausted response. Please reconcile the middleware with the fixes now in main and verify its ordering and body handling. Test complete, readable large compressed errors alongside successful and small uncompressed responses before approval.

Maneesh-Kumar-Maddini:
Thanks, Jayden. The update prevents the compressed body from being consumed, but returning encoded errors unchanged also skips standardisation for legacy JSON error responses. The new large-error test starts with `error_response()`, so it does not cover that case. Please add a large non-standard JSON error test with gzip enabled and verify both the complete readable body and the standard error envelope. Resolve the `main.py` conflict and rerun the compression/error checks before approval.

PR #1019, opened 2026-09-06
B3.1: Add cursor pagination for GET /detections/cursor

HarryTrieu:
## Summary
- Add `GET /detections/cursor` with keyset pagination using `timestamp` + `_id`.
- Return `items`, `limit`, and `next_cursor` while keeping existing `page`/`page_size` list behaviour unchanged.
- Reject malformed cursors with HTTP 400.

## Test plan
- [ ] Seed 25+ detections via `POST /detections`
- [ ] `GET /detections/cursor?limit=10` returns 10 items and a `next_cursor`
- [ ] Paste `next_cursor` for page 2 and confirm no duplicate `_id`s
- [ ] Walk pages until `next_cursor` is null
- [ ] `GET /detections?page=1&page_size=10` still returns page fields
- [ ] `GET /detections/cursor?cursor=not-valid` returns 400

### DIFF
--- a/src/production/backend/app/detections.py
+++ b/src/production/backend/app/detections.py
@@ -1,6 +1,7 @@
-from typing import List, Optional, Dict, Any
+import base64
+import json
+from typing import List, Optional, Dict, Any, Tuple
 from datetime import datetime
-from fastapi import HTTPException
 
 from bson import ObjectId
 from pymongo import ReturnDocument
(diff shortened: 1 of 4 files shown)


Maneesh-Kumar-Maddini (changes requested):
Cursor validation has two reproducible gaps: malformed ObjectIds raise InvalidId outside the current handling, while a null ID is accepted and generates a new ObjectId. Please require a valid, non-null cursor ID and return the intended client error for malformed cursors. Add tests for invalid encoding, malformed JSON, invalid/null IDs, duplicate timestamps, page boundaries, and the final page. Resolve the merge conflicts and verify pagination against current main before approval.

PR #1020, opened 2026-09-06
B1.2: Stream detections after MongoDB persistence

HarryTrieu:
## Summary
- Broadcast enriched event payloads over `/ws/detections` only after `Events.insert_one` succeeds.
- Add `DetectionStreamManager` and register the live WebSocket route in the API.
- Require a valid JWT (`?token=` or `Authorization: Bearer`) before accepting `/ws/detections`.
- Keep Mongo insert/lookup off the event loop via `asyncio.to_thread`, and wrap broadcast in try/except so a stream failure cannot turn a successful persist into a 500.
- Connect the HMI map to live detections via a WebSocket client (uses `localStorage.token`, `window.ECHO_API_WS_URL` for deploy).

## Test plan
- [x] Rebased onto latest `main` (resolved `main.py` conflict: keep both `payments` + `live` routers)
- [x] `/ws/detections` rejects missing/invalid token (WS close 1008)
- [x] `/ws/detections` accepts valid JWT from login `localStorage.token`
- [x] `POST /engine/event` still returns 201 after persist even if broadcast/enrichment fails
- [x] Mongo work in `create_event` runs via `asyncio.to_thread` (non-blocking for open sockets)
- [x] Manual Docker/Swagger/HMI persist-then-stream path previously verified (save -> WS toast/console). Re-run after Docker is up: login HMI first so token exists, then `POST /engine/event`
- [x] Deploy note: set `window.ECHO_API_WS_URL` (e.g. `wss://api.example.com`) outside local Docker; localhost fallback is local-only

### DIFF
--- a/src/production/backend/app/routers/engine.py
+++ b/src/production/backend/app/routers/engine.py
@@ -3,7 +3,10 @@ from fastapi import status, APIRouter
 from app import serializers
 from app import schemas
 from app.database import Events
+from app.services.detection_stream import detection_stream_manager
+import asyncio
 import datetime
+import logging
 from app import serializers
 from app import schemas
 from app.database import Events, Species
@@ -12,16 +15,60 @@ from fastapi.responses import StreamingResponse
 import pandas as pd
 
 
+logger = logging.getLogger(__name__)
 router = APIRouter()
 
-@router.post("/event", status_code=status.HTTP_201_CREATED)
-def create_event(event: schemas.EventSchema):
 
-    result = Events.insert_one(event.dict())
+def _build_created_event(inserted_id):
     pipeline = [
-            {'$match': {'_id': result.inserted_id}},
-        ]
-    new_post = serializers.eventListEntity(Events.aggregate(pipeline))[0]
+        {"$match": {"_id": inserted_id}},
+    ]
+    return serializers.eventListEntity(Events.aggregate(pipeline))[0]
+
+
+def _build_stream_payload(inserted_id):
+    stream_pipeline = [
+        {"$match": {"_id": inserted_id}},
+        {
+            "$lookup": {
+                "from": "species",
+                "localField": "species",
+                "foreignField": "_id",
+                "as": "info",
+            }
+        },
+        {
+            "$replaceRoot": {
+                "newRoot": {
+                    "$mergeObjects": [{"$arrayElemAt": ["$info", 0]}, "$$ROOT"]
+                }
+            }
+        },
+    ]
+    return serializers.eventSpeciesListEntity(
+        Events.aggregate(stream_pipeline)
+    )[0]
+
+
+@router.post("/event", status_code=status.HTTP_201_CREATED)
+async def create_event(event: schemas.EventSchema):
+    # Keep Mongo work off the event loop so open WebSocket clients stay responsive.
+    result = await asyncio.to_thread(Events.insert_one, event.dict())
+    new_post = await asyncio.to_thread(_build_created_event, result.inserted_id)
+
+    # Persistence already succeeded. Broadcast failures must not turn this into a 500.
+    try:
+        stream_payload = await asyncio.to_thread(
+            _build_stream_payload, result.inserted_id
+        )
+        await detection_stream_manager.broadcast(stream_payload)
+    except Exception:
+        logger.warning(
+            "Event %s persisted but live broadcast failed",
+            result.inserted_id,
+            exc_info=True,
+        )
+
     return new_post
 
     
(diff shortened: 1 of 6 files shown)


src/production/backend/app/routers/live.py
```diff
@@ -8,20 +9,15 @@ async def websocket_test(websocket: WebSocket):
...
+            await websocket.receive_text()
     except WebSocketDisconnect:
         pass
 
-# --- WebSocket endpoint for real-time detections ---
 @router.websocket("/ws/detections")
-async def websocket_endpoint(websocket: WebSocket):
-    await websocket.accept()
+async def detection_stream(websocket: WebSocket):
+    await detection_stream_manager.connect(websocket)
```
Vamshi-Gollapelly, line 42: This endpoint doesn't check for a session before accepting the connection. Anyone who can hit /ws/detections gets a live feed of species, sensor id, confidence and lat/lon for every detection, no auth needed.

Given this is wildlife location data, that feels like a real risk (poaching etc), similar to the PII issue on #1008. Can we either add a session check here, or get explicit sign-off that we're deferring auth on this with a follow-up issue?

src/production/backend/app/routers/engine.py
```diff
@@ -15,13 +16,36 @@
 router = APIRouter()
 
 @router.post("/event", status_code=status.HTTP_201_CREATED)
-def create_event(event: schemas.EventSchema):
-
+async def create_event(event: schemas.EventSchema):
```
Vamshi-Gollapelly: This is now async def but insert_one and aggregate are still plain pymongo calls (sync/blocking). Once it's async, FastAPI runs it directly on the event loop instead of offloading to a thread pool like it would for a regular def route. So while Mongo is doing its thing here, it blocks everything else on the event loop too, including the open /ws/detections connections.

Might be worth either keeping this as def and pushing the broadcast off some other way, or switching these two calls to an async driver (motor) if we want to keep it async. Not blocking on its own, but flagging since it could cause the websocket to lag/stall under load.

src/production/backend/app/routers/engine.py
```diff
@@ -15,13 +16,36 @@
...
+        },
+        {
+            "$replaceRoot": {
+                "newRoot": {
+                    "$mergeObjects": [{"$arrayElemAt": ["$info", 0]}, "$$ROOT"]
+                }
+            }
+        },
+    ]
+    stream_payload = serializers.eventSpeciesListEntity(
```
Vamshi-Gollapelly: Small one: if anything in here throws after the insert (e.g. this aggregate comes back empty and the [0] indexing blows up), the event is already saved in Mongo but the client gets a 500 and never sees it come back. So it'd look like the create failed when it actually didn't.

Might be worth wrapping the stream_payload build + broadcast in a try/except that just logs a warning on failure, since the write itself already succeeded by this point and shouldn't be undone by a broadcast problem.

Vamshi-Gollapelly (dismissed):
Also, not blocking on their own, but worth sorting before merge:

- Branch shows merge conflicts with main now, needs a rebase and a test run before this goes in
- Test plan checkboxes in the description are all still unchecked, were the manual steps (Docker stack, WS connect, POST /engine/event, etc.) actually run before opening this?
- detection_stream_client.js falls back to ws://localhost:9000 if window.ECHO_API_WS_URL isn't set, fine for dev, just confirm that's wired up wherever this deploys

Rest of it looks solid, the persist-then-broadcast approach is the right call. Happy to re-review once the auth and conflict points are handled.

HarryTrieu:
Thanks for the review @Vamshi-Gollapelly — addressed all points:

### Auth on `/ws/detections`
Added a session/JWT check before accept. Clients must pass a valid token via `?token=` (HMI uses `localStorage.token` from login) or `Authorization: Bearer`. Missing/invalid tokens are closed with WS 1008. This keeps wildlife location data off the open internet.

### async + sync pymongo
`create_event` stays async for `await broadcast`, but `insert_one` / aggregate work now runs in `asyncio.to_thread(...)` so the event loop (and open WS clients) are not blocked on Mongo.

### Persist success vs broadcast failure
Stream payload build + broadcast are wrapped in try/except. If enrichment/broadcast fails after insert, we log a warning and still return the 201 create response.

### Merge conflicts
Rebased onto latest `main` and kept both `payments` and `live` router includes.

### Test plan / localhost WS
Updated the PR checklist. HMI WS client now:
- requires login token
- uses `window.ECHO_API_WS_URL` when set
- falls back to `ws://localhost:9000` only on localhost
- otherwise derives `ws(s)://` from the page host for same-host deploys

Happy to re-review.

KishanFW (approved):

PR #1025, opened 2026-09-07
Redis Query Cache with Invalidation (C2.1)

Jamie-Maintainer:


KCP17 (changes requested):
This is a solid cache implementation overall (correct key parameterisation, fails open if Redis is down), but just a few things before merging. First, `invalidate_insights()` is defined but is never called from any write path, which means `/insights/*` can now serve up to 5 minutes of stale data after a new detection, which is a real regression this PR introduces, not just a documented gap. Can this get wired into the detection/event write paths as part of C2.1, rather than deferred? Second, the `docker-compose.yml` changes adding hardcoded Mongo credentials (`MONGODB_URI`/`MONGO_URI` with `root:root_password`) look unrelated to Redis caching. Can you clarify why they're here, or should they be a separate PR? Third, `prometheus-client` is added to `requirements.txt` but doesn't appear to be used anywhere in this diff. Is that for follow-up work? Also, a PR description would help a lot here given the scope of what's bundled in.

Jamie-Maintainer:
> This is a solid cache implementation overall (correct key parameterisation, fails open if Redis is down), but just a few things before merging. First, `invalidate_insights()` is defined but is never called from any write path, which means `/insights/*` can now serve up to 5 minutes of stale data after a new detection, which is a real regression this PR introduces, not just a documented gap. Can this get wired into the detection/event write paths as part of C2.1, rather than deferred? Second, the `docker-compose.yml` changes adding hardcoded Mongo credentials (`MONGODB_URI`/`MONGO_URI` with `root:root_password`) look unrelated to Redis caching. Can you clarify why they're here, or should they be a separate PR? Third, `prometheus-client` is added to `requirements.txt` but doesn't appear to be used anywhere in this diff. Is that for follow-up work? Also, a PR description would help a lot here given the scope of what's bundled in.

**Scope:**
- Added Redis-backed caching for selected /insights queries.
- Added parameterised cache keys for overview/species queries.
- Configured Redis DB 2 and finite TTL.
- Added fail-open Redis cache access so API requests continue if Redis is unavailable.
- Added cache invalidation after relevant detection/event write operations.

**Verification**
- Verified cache miss followed by cache hit.
- Verified cache keys are stored in Redis DB 2.
- Verified TTL expiry configuration.
- Verified detection/event writes invalidate affected insight cache entries.
- Compared cache miss and cache hit latency.

**Scope cleanup**
- Removed unrelated MongoDB credential/configuration changes.
- Removed unrelated Prometheus dependency changes from this PR.

Please check the deliverables for more details:
[Sprint 2 C2.1 Deliverable (Jamie).pdf](https://deakin365.sharepoint.com/:b:/s/DataBytes2/IQA3kUnBpXyKTKM-_nKrEdo4AUB1FE6JudgL3qoRTFoSYd8?e=5Qu1d0)

KCP17 (approved):
I rereviewed the current diff and confirmed all three original issues are fixed: invalidate_insights() is now called in detections.py (create/update/delete) and routers/engine.py (create_event), which are the only two places anywhere in the backend that write to Events, I checked every other insert_one/update_one/delete_one call in the app and confirmed they write to different collections (AudioUploads, Predictions), so there's no remaining stale-cache gap. The unrelated docker-compose.yml Mongo credential changes and the unused prometheus-client dependency are both gone from this diff, and the PR description now has a proper scope/verification writeup. I'm approving, good work!

PR #1027, opened 2026-09-10
FR-D4: reCAPTCHA and security hardening (login flow)

Mustafa-1089:
Wires real, env-backed reCAPTCHA into the login flow. Scoped to login only for a clean review.

## Changes
- Site key now loads from `RECAPTCHA_SITE_KEY` via a new `/api/config/recaptcha` endpoint, no key hardcoded in HTML
- Login form captures the captcha token, blocks submit if missing (when enabled), and forwards it through the HMI proxy to the backend
- Added expired / error / unavailable states and widget reset on failed login
- `api.js` set to explicit render mode to avoid an auto-scan error when captcha is disabled
- CSP for reCAPTCHA hosts already in place

## Fails safe
With no key set, the widget hides and login works normally. Verified locally: `/api/config/recaptcha` returns `{"enabled":false,"siteKey":""}`, widget hidden, login succeeds, no hardcoded key in source, console clean of the sitekey error.

## Backend handoff (follow-up)
`auth.routes.js` forwards `captchaToken` to `/hmi/signin`. The backend still needs to verify it against Google's `siteverify` and reject invalid tokens; Backend team's half.

## To activate
Register a reCAPTCHA v2 key, then set `RECAPTCHA_SITE_KEY` (HMI) and `RECAPTCHA_SECRET_KEY` (backend).

## Notes
- Token-send and expiry paths verified by code review, pending a real key for runtime testing
- Pre-existing commented-out keys and old `VerifyCaptcha` button left untouched

### DIFF
--- a/src/production/hmi/ui/public/admin/sensor_health/script.js
+++ b/src/production/hmi/ui/public/admin/sensor_health/script.js
@@ -6,6 +6,16 @@
 // setInterval timer and the interval shown to the user, so they cannot drift.
 const REFRESH_MS = 15000;
 
+// Small muted style for the "seen X ago" line under each status pill.
+(function injectSensorHealthStyles() {
+  if (document.getElementById("sensor-health-inline-styles")) return;
+  const style = document.createElement("style");
+  style.id = "sensor-health-inline-styles";
+  style.textContent =
+    ".cell-subtext{font-size:11px;color:var(--muted,#888);margin-top:2px;}";
+  document.head.appendChild(style);
+})();
+
 const menuToggle = document.getElementById("menu-toggle");
 const mobileBackdrop = document.getElementById("mobile-backdrop");
 
@@ -173,6 +183,22 @@ function formatUptime(seconds) {
   if (days > 0) return `${days}d ${hours}h ${mins}m`;
   if (hours > 0) return `${hours}h ${mins}m`;
   return `${mins}m`;
+}
+
+// Turn a "minutes ago" number into readable text for last-seen / last-audio.
+function formatMinutesAgo(minutes) {
+  if (minutes === null || minutes === undefined || minutes === "") return "—";
+  const num = Number(minutes);
+  if (!Number.isFinite(num) || num < 0) return "—";
+  if (num === 0) return "just now";
+  if (num < 60) return `${num}m ago`;
+
+  const hours = Math.floor(num / 60);
+  const mins = num % 60;
+  if (hours < 24) return mins ? `${hours}h ${mins}m ago` : `${hours}h ago`;
+
+  const days = Math.floor(hours / 24);
+  return `${days}d ago`;
 }
 
 function updateLastUpdated() {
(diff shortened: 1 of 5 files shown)


src/production/hmi/ui/public/admin/sensor_health/script.js
```diff
@@ -165,6 +209,51 @@ function updateLastUpdated() {
...
+        counts.degraded += 1;
+        break;
+      case "Offline":
+        counts.offline += 1;
+        break;
+      case "Low Battery":
+        counts.lowBattery += 1;
+        break;
+      default:
+        counts.other += 1;
```
Vamshi-Gollapelly, line 238: counts.other is tracked but never shown on a KPI card once sensors can report Unknown (from #1002), the total won't match the sum of the visible cards, with nothing explaining the gap. Worth either a fifth KPI card for it or folding it into an existing one and adding "Unknown" to the status filter dropdown too.

Vamshi-Gollapelly (changes requested):
Nice, clean implementation of the reCAPTCHA wiring — fails open sensibly when no key is set, and the callback states (expired/error) are a good touch.

Two things before this is ready to merge:

The description says this is "scoped to login only," but the diff also changes sensor-health.html and sensor_health/script.js (KPI cards, status pill/filter logic) unrelated to login/reCAPTCHA, and one comment even references "FR-C1." Could this be split into two PRs?
Since backend verification of captchaToken against Google's siteverify is called out as a follow-up, merging this alone doesn't add real bot protection yet just a UI gate. Can you link the backend PR here so it's tracked?     
 
One more this PR and #1002 both rewrite sensor-health.html and sensor_health/script.js independently. #1002 adds Type/Model/Temp/Components/Details columns to the table; this PR's row template only builds 9 columns and doesn't include them. Whichever of these merges second will need a real rebase against the other, not just conflict resolution otherwise the Details link or this KPI work will get lost. Might be worth you and Balasaiteja syncing before either merges.

src/production/hmi/ui/public/login.html
```diff
@@ -304,10 +306,85 @@
...
+                    if (host) { host.style.display = 'none'; }
+                    _setLoginSubmitEnabled(true);
+                    return;
+                }
+
+                window.__captchaEnabled = true;
+                _setLoginSubmitEnabled(false);
+
+                // grecaptcha may still be loading (api.js is async/defer)
+                function renderWhenReady() {
```
AnDo081105, line 368: A missing Google script causes endless polling and a permanently disabled login. Authentication failures return HTML, so JSON parsing fails before the new CAPTCHA-reset branch runs.

AnDo081105 (commented):
MFA’s required JSON contract remains broken. [auth.routes.js (line 71)](/D:/Study/Y3T2/SIT378/proj/Project-Echo/src/production/hmi/ui/routes/auth.routes.js:71) returns HTML-with-200 for MFA, while [login.html (line 496)](/D:/Study/Y3T2/SIT378/proj/Project-Echo/src/production/hmi/ui/public/login.html:496) parses every successful response as JSON. MFA login therefore throws instead of opening OTP verification.

PR #1030, opened 2026-09-11
Integrate admin dashboard with live API data

KavinduLiyanage04:
## Summary
- Added a detection summary endpoint for the admin dashboard
- Connected the HMI dashboard through API proxy routes
- Replaced hardcoded chart values with live detection and donation data
- Added loading, error and empty-data states

## Testing
- Tested the API directly through port 9000
- Tested the HMI proxy through port 3000
- Confirmed the dashboard loads successfully
- The local detections collection is empty, so the chart correctly displays no detection data


KishanFW (approved):
All good

PR #1032, opened 2026-09-11
B2.3: Add analytics filters to insights endpoints

HarryTrieu:
﻿## Summary
- Extend `GET /insights/overview` and `GET /insights/species` with shared date, species, and sensorId filters.
- Use Mongo aggregation for overview counts instead of loading all events into Python.
- Return HTTP 400 for invalid start/end datetime query params.
- Keep unfiltered behaviour backward compatible.
- Normalize mixed Mongo timestamp formats (`Date` and ISO strings) before response formatting so insights endpoints do not crash on `.isoformat()`.

## Test plan
- [x] `GET /insights/overview` with no filters still returns counts
- [x] Filter by `species`, `sensorId`, and date range against seeded events
- [x] Combined filters return intersection counts
- [x] `GET /insights/species?sensorId=...` returns filtered species breakdown
- [x] Invalid `start=not-a-date` returns 400
- [x] Re-run quick Swagger/HTTP smoke check after review
- [x] Response formatting handles datetime values and existing string timestamps without crashing
- [x] Filtered insights requests covered for both datetime and string timestamp records
- [x] Unsupported timestamp shapes are skipped instead of failing the whole request

### DIFF
--- a/src/production/backend/app/routers/insights.py
+++ b/src/production/backend/app/routers/insights.py
@@ -1,80 +1,163 @@
-import os
-from fastapi import APIRouter, Query
 from datetime import datetime
-from typing import Optional
-import pymongo
+from typing import Any, Dict, Optional
+
+from fastapi import APIRouter, HTTPException, Query
+
+from app.database import Events, Microphones, Nodes
 
 router = APIRouter(prefix="/insights", tags=["insights"])
 
-MONGO_URI = os.getenv("MONGO_URI")
-MONGO_DB = os.getenv("MONGO_DB", "EchoNet")
 
-client = pymongo.MongoClient(MONGO_URI)
-db = client[MONGO_DB]
-
-
-def parse_ts(value):
+def _parse_ts(value: Any) -> Optional[datetime]:
     if isinstance(value, datetime):
         return value
     if isinstance(value, str):
         try:
             return datetime.fromisoformat(value.replace("Z", "+00:00"))
-        except Exception:
+        except ValueError:
             return None
     return None
 
 
+def _parse_query_ts(value: Optional[str], field_name: str) -> Optional[datetime]:
+    if value is None:
+        return None
+
+    parsed = _parse_ts(value)
+    if parsed is None:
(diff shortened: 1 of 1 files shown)


Maneesh-Kumar-Maddini (changes requested):
The analytics filters are implemented, but the response-building code assumes stored timestamps always provide .isoformat(). Existing records containing string timestamps can therefore cause the endpoint to fail.
Please normalize supported timestamp formats before formatting the response, or handle unsupported records explicitly without crashing the entire request. Add coverage using both datetime values and existing string timestamps, including filtered requests, and confirm the insights endpoints return valid responses for both.

HarryTrieu:
﻿Addressed the timestamp formatting review feedback in e9087c34:

**Fix**
- Insights aggregation now normalizes mixed Mongo `timestamp` storage (`Date` and ISO/Zulu strings) via `$dateFromString` before filtering/grouping.
- Response `timeRange` uses `_to_iso_string()` so datetime **and** string values are formatted safely; unsupported records are skipped instead of crashing the request with `AttributeError` on `.isoformat()`.

**Coverage**
- Added `tests/test_insights_timestamp_normalization.py` covering datetime values, existing string timestamps, unsupported shapes, and filtered `/insights/overview` + `/insights/species` requests.
- Verified locally against live Mongo with mixed seed docs (`B2.3-TS-A` datetime, `B2.3-TS-B` string, `B2.3-TS-C` unsupported) — filtered overview/species return valid JSON for both supported formats; unsupported docs do not fail the endpoint.

Ready for re-review.

Maneesh-Kumar-Maddini (approved):
Thanks for addressing the timestamp-handling feedback in e9087c34. I reviewed the updated implementation and regression coverage.

The aggregation now normalizes datetime and string timestamps before filtering and grouping, safely skips unsupported records, and formats the response without the previous .isoformat() failure. The four new regression tests passed in isolation, and the latest build check has passed.

This resolves my previous review concern, and I’m happy to approve this scoped change.

PR #1041, opened 2026-09-14
Wire ENGINE_API_KEY into docker-compose for echo_engine and echo_api

AshanVimodh:
Krish confirmed both sides are ready (PR #1037 already sends X-Engine-Api-Key when ENGINE_API_KEY is set) and asked for a shared secret via a gitignored .env rather than hardcoding it in docker-compose.yml. Pass it through Compose's existing .env variable-substitution (src/deployment/docker/.env, already used for Mongo/JWT/Twilio secrets) rather than env_file, so a missing value degrades to empty/unauthenticated instead of failing docker compose up for anyone who hasn't set one.

Verified against the real containers: built and ran echo_store + echo_api from the actual Dockerfile with a generated key in .env. Request with no header -> 401, wrong key -> 401, correct X-Engine-Api-Key -> 201, and the inserted MongoDB event matches the agreed contract exactly. Did not build/run echo_engine itself (large CUDA-based image); Engine-side behavior is confirmed by inspection of the code merged in PR #1037.


KishanFW (approved):
All good

PR #1042, opened 2026-09-14
Superseded by #1043

25qdatttt:
# Testing Framework Research and TDD Guide

**Required Work**
Research and practically test one or two testing frameworks (pytest, Jest, Locust, k6, Artillery) covering unit, integration/e2e, and load/performance testing, document findings, and produce a short TDD guide for future students, per Krish's request and his follow-up narrowing load testing to Locust and k6 specifically.

Set up `pytest` to unify Backend, Engine, and HMI testing, absorbing Engine's existing 44 tests with zero changes. Practically ran and compared both `Locust` and `k6` for load testing, then pushed both tools across 7 escalating levels (10 to 1000 concurrent users) to find the backend's actual breaking point, instead of just confirming a comfortable load level.

This surfaced 4 real bugs, one fixed here:

- **Fixed**: a cross-test-pollution bug in `test_iot_integration.py`, where an unclean `sklearn` module stub broke `test_heldout_baseline.py` whenever the full suite ran together. Fixed by restoring `sys.modules` right after the import that needed the mock.
- **Found, not fixed** (out of scope, documented for the owning area): `GET /insights/overview` and `/insights/species` return 500 in any real deployment because `app/routers/insights.py` reads the wrong env var name (`MONGO_URI` instead of `MONGODB_URI`), with no fallback.
- **Found, not fixed**: `app/main.py` crashes on startup outside a real interactive terminal, because it prints a checkmark character that Windows' default console codepage can't encode.
- **Found, not fixed**: a live Gmail app password is committed in plaintext in `app/routers/sim.py`.

Ran the load-testing comparison against 4 verified-safe endpoints. At a 20-user baseline, both tools agree closely (about 0% failures, ~5ms median). Pushed further to 1000 users: the backend stays clean up to ~200 users, then a real capacity ceiling appears from ~300 users on, caused by a single uvicorn worker's connection backlog. `k6` and `Locust` report that same ceiling in completely different ways: k6 surfaces it as explicit connection failures (1.88% rising to 5.36%), while Locust shows 0.00% failures throughout and only its latency (median jumping to 130ms) reveals the same problem.

**Expected Deliverables**

- **Framework selection and setup**: done. `pytest` for unit/integration/e2e across Backend, Engine, and HMI; `Locust` and `k6` for load testing.
- **Practical validation**: done, every command in the guide was actually run against the real codebase (61 Engine tests, 6 Backend tests, 19 HMI tests, plus both load-test runs, all passing).
- **TDD guide for future students**: done, see `docs/team-guides/TDD_Guide.md`.

Full technical log, per-level load numbers, and root-cause analysis in `Testing_Framework_Research_Report_Nolan_Nguyen.md` (repo root, submitted separately per team convention, not committed to git).


25qdatttt:
Closing to recreate manually.

PR #1044, opened 2026-09-15
EE/VHL: fix BackendAdapter LLA format

Mangon3:
Restores object-form LLA in the Engine to Backend payload. Fixes the 3 end-to-end test failures that have been on main since 14 September. Engine suite: 82/83 locally (1 error is diskcache missing in my venv).


Maneesh-Kumar-Maddini:
The location-format fix looks correct. `diskcache` is already declared in the engine requirements, so no dependency code change is requested. Please install the existing dependencies in the environment used to run the tests, rerun the complete engine suite, and update the PR with the results before final approval.

KishanFW (approved):
All good

PR #1046, opened 2026-09-16
EE/QDN: Similar-Detection Retrieval (Embedding-Based Detection Review)

25qdatttt:
Adds GET /detections/{id}/similar, which ranks past detections by cosine similarity between the model's own embeddings, plus an admin page (Detection Review) so reviewers can compare a detection's audio against its closest historical matches instead of trusting confidence alone.

Also fixes two bugs found along the way:
- The embedding space was severely collapsed (97% dead dimensions), traced to frozen BatchNorm plus training without pretrained weights. Retrained with both fixed: 0% dead dimensions, 97.9% nearest-neighbour accuracy.
- GET /detections returned 500 on real Mongo data due to a missing ObjectId encoder in DetectionListResponses.

27 tests passing, zero regressions. Full investigation and numbers in Sprint2_HD_Feature_Report_Nolan_Nguyen.md (repo root, not committed, submitted separately). Usage guide: docs/team-guides/Detection_Review_User_Guide.md.


Maneesh-Kumar-Maddini (changes requested):
Similar-detection retrieval has a reproducible failure when candidate embeddings have different vector lengths: the similarity helper raises an uncaught ValueError. Please validate dimensions and finite values, and handle incompatible candidates without failing the entire request. Add regression coverage for malformed and mixed-dimension embeddings. Also attach the referenced model evaluation report and reproduction details, because the separately submitted model results cannot be verified from the available PR contents.

25qdatttt:
> Similar-detection retrieval has a reproducible failure when candidate embeddings have different vector lengths: the similarity helper raises an uncaught ValueError. Please validate dimensions and finite values, and handle incompatible candidates without failing the entire request. Add regression coverage for malformed and mixed-dimension embeddings. Also attach the referenced model evaluation report and reproduction details, because the separately submitted model results cannot be verified from the available PR contents.

Fixed in commit c70398c3.

top_k_similar() now skips any candidate with a missing, non-finite (NaN/Inf), or dimension-mismatched embedding instead of letting the ValueError propagate and fail the whole request. Each skip is logged with the reason, so a recurring data-quality issue (e.g. detections stored by an old model version) stays visible instead of silently swallowed.
The queried detection's own embedding is validated separately and now returns a clear 422 if malformed, since that case can't be skipped the way a candidate can.
Added regression tests for both cases: mixed-dimension candidates, NaN/Inf values, and the boolean edge case (bool is a subclass of int in Python).
Full suite: 28 passed, no regressions.
Attaching the model evaluation report below with the embedding-collapse investigation, retraining comparison, and reproduction steps.
[Sprint2_HD_Feature_Report_Nolan_Nguyen.pdf](https://github.com/user-attachments/files/32410724/Sprint2_HD_Feature_Report_Nolan_Nguyen.pdf)

Maneesh-Kumar-Maddini (changes requested):
Reviewed commit 6c7650f and the attached evaluation report. The earlier embedding-validation issue is resolved: incompatible or non-finite candidates are skipped, and malformed query embeddings return 422. All 28 backend tests run in isolation passed, and all four CI checks are green.

Two issues remain before approval:

1. Restore the Detection Review page’s connection to the backend. The page requests /detections, /detections/{id}, and /detections/{id}/similar, but the current HMI server no longer forwards those paths. Route-level checks return 404. Add authenticated forwarding for the required endpoints, or update the page to supported routes with the correct response format. Verify listing, selection, similarity results, and audio through the HMI.

2. Render stored species names safely. The list, detail heading, and similarity rows interpolate species directly into innerHTML. The backend accepts and returns markup in this field, so stored content can become executable HTML. Use textContent or appropriate escaping and add regression coverage showing that markup is displayed as text.

The previous review points are addressed, but approval remains pending these HMI fixes and verification of the updated revision.

Vamshi-Gollapelly (approved):
LGTM. Went through both rounds of requested changes and they're actually fixed, not just patched over:

The embedding-validation crash is gone bad candidates (wrong dimension, NaN/Inf) get skipped and logged instead of blowing up the whole request, and there's a 422 now if the query detection's own embedding is bad. Tests cover the mixed-dimension and NaN cases plus the bool-is-secretly-an-int gotcha.
HMI connectivity and the XSS issue are both handled the proxy routes are back, and species names go through escapeHtml everywhere before hitting innerHTML now. Tried to think of a way around it and couldn't.

Approving. One small thing for later, not blocking: the "weak match" fade cutoff (0.6) and the "novel" badge cutoff (0.5) don't match, so you can get a result badged "Consistent match" that's also faded as "weak, not a strong precedent" in the same view. Might be worth aligning those or adding a line to the doc about why they're different.

Maneesh-Kumar-Maddini (approved):
Reviewed commit cad7db2. The previously requested fixes are complete: the Detection Review page now uses authenticated HMI forwarding routes, and stored species names are safely escaped throughout the page. The embedding-validation fixes remain intact.

Verification passed with 28 backend tests, 17 HMI tests, and an isolated page-to-proxy check covering detection listing, similarity results, candidate audio decoding, and safe rendering. All four GitHub CI checks pass, and the branch is current with main.

All review points have been addressed. This PR is ready to merge.

PR #1049, opened 2026-09-16
EH/KT/FR-A4 Live Map Controls and Filters

kimmynoto:
Adds pause/resume, manual refresh, live "last updated" timestamp, and two working filters (event type, species) to the live map.

**Filters implemented:**
- Event type (vocalization, movement, sensor_health, iot_node)
- Species (real data for movement events; not populated for vocalization events sourced from raw MQTT, since those carry a placeholder value until DB enrichment)

**Not implemented, confirmed with team lead:**
- Conservation status, no real value currently available in the live MQTT event schema; the existing `status` field is either a placeholder or represents device health depending on event type, not conservation status
- Device type, field doesn't exist in the current event payload

Both are documented here rather than built with fake/hardcoded data. Team lead confirmed the live event contract doesn't currently support these two, and prioritized the main integration flow (audio → Engine → Backend/MongoDB → HMI) and `sourceType` (Simulator vs Real) distinction for this sprint instead. These can be picked up separately if enrichment/device metadata is added later.

**Note:** filters apply to new incoming events only, not markers already on the map at the time a filter is changed.


KCP17 (approved):
First, looks like this also fixes a real duplicate-code bug in `_addVocalizationFeature` (the setStyle/setId/addFeature block was repeated twice in the base file). Was that intentional cleanup? Second, there are two leftover debug comments (near the `audio_recorder.js` import and the `audioRecorder` variable) that claim something was commented out or stubbed, but the actual lines below them are still active and unchanged. Is there a real pre-existing syntax error in `audio_recorder.js` worth knowing about, or can these dead comments just be removed?

PR #1054, opened 2026-09-17
A2.2: Document audio clip metadata verification and budget-check bug …

mansipatel765-sk:
## Task
A2.2 (Shared, under A2 – Audio Clip Storage and Detection Linking)

## Summary
Verified audio upload → detection linkage flow end-to-end using the existing POST /api/audio/upload route. Documented metadata/index requirements. Found two separate budget-related blockers: (1) POST /detections returns 403 because no budget rule is configured for "detections", a config gap fixable via POST /admin/budget/limits; (2) GET /api/admin/budget/usage returns 500 due to a get_usage() / month_key bug in admin_budget.py, which is independent of the 403.

## Evidence
See `src/production/backend/docs/sprint2/A2.2-audio-clip-metadata-verification.md`
for full test steps, requests/responses, and findings.

## Reviewer
As per Sprint 2 allocation (A2, Shared)


KCP17 (changes requested):
`enforce_and_consume()` (the dependency actually guarding `POST /detections`) never calls `get_usage()` at all, it does its own separate `limit <= 0` check. Thus the `403` isn't caused by the `get_usage`/`month_key` bug, but is caused by there being no budget rule configured yet for `"detections"`, a config gap fixable via `POST /admin/budget/limits`, independent of the `get_usage()` bug. Worth splitting these into two separate blockers in the writeup so whoever picks this up doesn't wait on a code fix that wouldn't actually unblock detection creation.

mansipatel765-sk:
Hi, thanks for the feedback. I've updated the writeup and the PR summary. The 403 on POST /detections and the get_usage() / month_key 500 are now split into two separate blockers. The 403 is a config gap (no budget rule for "detections", fixable via POST /admin/budget/limits) and the 500 is an independent code bug. Please have another look when you get a chance.

PR #1055, opened 2026-09-17
C2.2: Document Redis cache invalidation and performance verification

mansipatel765-sk:
## Task
C2.2 (Shared, under C2 – Redis Query Cache with Invalidation)

## Summary
Verified the Redis query cache built for C2.1 against /insights/species. Cache miss/hit flow works as designed (76ms uncached vs 6ms cached in one run). Re-tested invalidation through the normal write path: after a POST /engine/event with a new species, the cached response stayed stale and the key was not cleared, and the new species only appeared after a manual cache flush. Separately, invalidate_insights() exists in app/cache.py but is never called from any write path.

## Evidence
See `src/production/backend/docs/sprint2/C2.2-redis-cache-invalidation-verification.md`
for full test steps, requests/responses, and findings.

## Reviewer
As per Sprint 2 allocation (C2, Shared) — no reviewer added per team instruction.


Maneesh-Kumar-Maddini (changes requested):
The cache experiment does not establish the claimed application-write invalidation behavior. It writes directly to detections, while the relevant insights aggregation reads events; direct database writes also bypass application invalidation hooks. Please repeat the experiment through the normal write path using data that changes the endpoint result. Compare the uncached result, warmed cache, and response after the write. Keep the unused-invalidation-helper code finding separate from the experimental conclusion.

mansipatel765-sk:
Hi, thanks for the feedback. I've redone the invalidation test through the normal write path. I warmed the cache on /insights/species, then wrote a new species via POST /engine/event (which writes to events, the collection the endpoint reads). The response stayed stale and the cache key wasn't cleared, and the new species only showed up after a manual cache flush. The writeup now compares the uncached, warmed and after-write results, and the unused invalidate_insights() finding is a separate section. Please have another look when you get a chance.

Maneesh-Kumar-Maddini:
Thanks, Mansi. The revised documentation addresses the review feedback by testing the normal write path and separating the observed stale-cache behaviour from the unused-helper finding. The comparison between uncached, warmed-cache, and after-write results clearly records the outcome. The documentation changes are ready for approval.

AshanVimodh (changes requested):
Thanks for reworking the methodology - testing through POST /engine/event against the events collection, and separating the unused-helper finding from the experimental conclusion, addresses Maneesh's earlier feedback directly.

Requesting changes on one point: the core finding is now out of date. C2.1 (#1025) merged on 2026-09-20 with a fix C2.1 issues commit that wired invalidate_insights() into engine.py's create_event - the exact write path this doc tests - plus three call sites in detections.py. That's after this doc was written (2026-09-17) against an earlier snapshot of the C2.1 branch, so the finding "invalidation after a write does not happen yet" / "invalidate_insights() is never called" no longer matches current main.

Could you re-run section 3 (uncached → warmed → after-write) against current main and update the findings/status/next-steps to match what actually happens now? If invalidation now works correctly, this becomes a much shorter confirmation doc rather than a bug report.

PR #1033, opened 2026-09-11
C14.2: Add consistency checker CLI and fixture validation

HarryTrieu:
## Summary
- Add report-only consistency checks for detection/event `audioClip` references to `audio_uploads`.
- Flag orphan `C14.2-SEED-*` audio uploads that are not referenced by any detection or event.
- Provide optional `--safe-fix --confirm` mode that only touches seeded fixture rows.
- Expose CLI as `python -m tools.check_consistency`.

## Test plan
- [x] Seed one valid upload + good detection link
- [x] Seed one broken detection (`audioClip` points to missing upload)
- [x] Seed one orphan upload with `C14.2-SEED-ORPHAN-01`
- [x] `--report-only` flags broken + orphan, not the good link
- [x] `--safe-fix` without `--confirm` refuses
- [x] `--safe-fix --confirm` cleans only `C14.2-SEED-*` fixtures
- [x] Re-run smoke check after review (branch includes current main; ordinary non-seed broken row left unchanged)

See latest PR comment for full CLI output from the post-merge smoke run.

PR #1057, opened 2026-09-17
C14.1: Add report-only database consistency checker with safe-fix mode

mansipatel765-sk:
## Task
C14.1 (Owner, under C14 – Database Consistency Checker)

## Summary
Implemented a standalone report-only consistency checker (tools/db_consistency_check.py) that detects detection/event records referencing an audioClip upload that doesn't exist. The scan reuses the shared checker from #1033 (run_consistency_checks()). Includes an optional conservative safe-fix mode that requires explicit --confirm. After review, the fix now resets a broken audioClip to an empty string (the app's existing "no audio clip" value, so the record stays valid against the string-field model) instead of null. The update also matches the original audioClip from the scan, so a link changed after scanning is skipped rather than overwritten, and repairs are verified by reading the record back.

## Evidence
See src/production/backend/docs/sprint2/C14.1-database-consistency-checker.md for the test steps: clean-state check, two deliberately broken records detected correctly, a repair that fixed one record and skipped one changed after the scan, verification results, and retrieval of the repaired detection through GET /detections/{id} (HTTP 200).

## Reviewer
As per Sprint 2 allocation (C14, Owner) — no reviewer added per team instruction.


Maneesh-Kumar-Maddini (changes requested):
The report-only checker is implemented, but its safe-fix behavior is not schema-safe. Setting audioClip to None makes a previously valid detection fail the existing model’s string-field validation; this was reproduced during review. Please use an application-supported representation and verify retrieval after repair. Updates should also match the original reference so they cannot overwrite a valid link changed after scanning. Reconcile this tool with the consistency checker already introduced through #1033.

mansipatel765-sk:
Hi, thanks for the review. I've made all three changes.

Repair value: the safe-fix now sets audioClip to an empty string (the value the app already uses for "no audio clip") instead of None, so the record stays valid against the string-field model. I checked retrieval after repair: GET /detections/{id} returns 200 with "audioClip": "".
Guarded update: the update now matches the original audioClip value from the scan, so a link changed after scanning is skipped, not overwritten. Tested with a record changed after the scan, and it was left as it was.
Reconciled with #1033: the scan now calls run_consistency_checks() from #1033, so there's one definition of a broken link. The repair stays in this tool, since the #1033 repair only touches seeded fixture records. The writeup explains this.

Please have another look when you get a chance.

Maneesh-Kumar-Maddini:
Thanks, Mansi. The updated implementation addresses the review findings. Repairs use the supported empty-string representation, updates check the original reference to avoid overwriting a changed link, and the tool reuses the consistency scan from #1033. The retrieval and concurrent-change verification cover the identified repair risks. These changes are ready for approval.

PR #1061, opened 2026-09-17
Lock down Donations API and add PII-free summary endpoint (Finding 6.7)

mansipatel765-sk:
## Summary
Addresses Finding 6.7 (Major) — PII exposure via unauthenticated donations endpoint.
Note: this endpoint lives in the Node.js HMI server (src/production/hmi/ui/server.js),
not the Python backend.

## Changes
- `GET /donations` now requires an admin session, using the existing session-based
  check (`verifyToken` + `isAdmin` from `middleware/authJwt.js`). The admin dashboard
  already authenticates with the session cookie, so `retrieveDonations()` and the
  page work unchanged.
- Added new public `GET /donations/summary` endpoint returning only aggregate
  totals (`totalRaised`, `goal`) — no PII.
- Fixed a pre-existing bug in the local `isAdmin` (still used by `/api/users/:id/...`)
  that crashed with a 500 when no Authorization header was sent; it now returns a clean 401.

## Verification
- Logged-in admin loads the donations dashboard: `GET /donations` returns 200 (screenshot in comments).
- `GET /donations` with no session returns 403 "No token provided!" and no donor data.
- `GET /donations/summary` returns aggregate-only JSON, no personal data.

## Related
Closes Finding 6.7 (Major) from App Attack team's vulnerability report.


Maneesh-Kumar-Maddini (changes requested):
The donation-detail protection does not match the existing authentication flow. The donation retrieval caller sends no bearer token, and the middleware checks a singular role field while backend JWTs contain a roles array. This can reject the normal admin workflow. Please align the caller and middleware with the established token/session contract. Verify that a real administrator succeeds, ordinary users are denied, missing credentials are rejected, and the public summary contains no donor-identifying information.

Vamshi-Gollapelly (changes requested):
Nice work closing the actual hole here isAdmin handles a missing header properly now instead of blowing up, /donations is locked down, and I dug into what's actually in that donations collection (name, email, paymentId were all just sitting there unauthenticated before, so good catch on that one). /donations/summary look clean too, just totals, nothing identifying.

But I think there's a snag that'll bite as soon as this merges. I traced through how the admin dashboard actually calls /donations it goes through retrieveDonations() in routes.js, and that function fires off api.get("/donations") with no auth header attached. I went looking for anything that might sneak a token in there automatically (interceptor, default header, anything) and came up empty in both routes.js and admin-layout.js. Compare that to something like retrieveUserNotifications, which explicitly takes a token and passes it along via authHeaders() donations never got that treatment. And donations.html itself doesn't reference localStorage or sessionStorage anywhere, so there's no JWT floating around client-side for it to grab even if we wanted to fix it that way. It's riding on the existing session cookie, not a bearer token.

So the way I read it, the first admin who opens that dashboard after these ships is going to get bounced with a 401. The verification section covers the "no token = rejected" case well, but I don't see anything confirming a logged-in admin can still actually see their donations. Could you either wire up how the JWT gets to the browser, or just swap isAdmin for whatever session-cookie check the rest of the admin pages already use?

mansipatel765-sk:
@Maneesh-Kumar-Maddini @Vamshi-Gollapelly thanks both. I've pushed f1406efc, which moves GET /donations from the header-only bearer check to the existing session-based check (verifyToken + isAdmin from middleware/authJwt.js). This matches how the dashboard already authenticates, so retrieveDonations() needs no change.

Verified locally:
- GET /donations with no session returns 403 "No token provided!" and no donor data.
- GET /donations/summary is unchanged and returns only totalRaised and goal.

Maneesh-Kumar-Maddini:
Thanks, Mansi. Using session authentication addresses the missing-header problem, but the imported middleware still needs to match the actual login contract. It reads `decoded.id` and checks `user.role === "Admin"`, while the backend authentication uses a different identity/roles contract. Please verify the complete flow with a real administrator login, confirm ordinary users are denied, and retain the public summary without donor details. Resolve the current `server.js` conflict and provide those results before approval.

PR #1064, opened 2026-09-17
Kd/fix map problem

AnDo081105:
# Summary

- fix the display of the map due to mistake in merge conflict resolve
- unify live map pause and refresh on the sidebar and popup


KishanFW (approved):
everything is good

PR #1066, opened 2026-09-18
EH/KD/chore

AnDo081105:
# Summary

- Update the HMI team codebase structure
- Update the HMI README.md


BalasaitejaG:
Location: server/app.js — line 161

With no session cookie and an empty mocked Redis store, GET /map and /index.html return 302 to /login, but GET /pages/map/index.html returns 200 with the map HTML. The new physical file lives underneath express.static and bypasses the guard. This confirms a page-access regression; it does not establish that every backend API becomes accessible.

BalasaitejaG:
Location:public/vendor/admin/styles.min.css
in line 9: tabler-icons.css,

The unchanged import ../css/icons/tabler-icons/tabler-icons.css now resolves to /vendor/css/icons/tabler-icons/tabler-icons.css (404). The actual moved file is /vendor/admin/tabler-icons/tabler-icons.css (200). Its old target existed in the base. Pages such as admin-nodes, projects, profile, donations and notifications rely on the theme import for Tabler icon definitions. The static-dependency test only parses url(...) and misses quoted @import statements. The Simplebar import on line 8 was already unresolved and is not counted as a new regression.

BalasaitejaG:
Location: public/features/audio/spectrogram.js

src/prototypes/hmi/audio_spectrogram_prototype/tests/production-spectrogram.test.js lines 19, 23 and 27 still read public/js/spectrogram.js, spectrogram-workflow.js and audio_recorder.js. Running npm test in that prototype gives 17 passing and 13 failing tests on the PR. Restoring the three exact base blobs in an isolated copy gives 32 passing and one pre-existing recorder/browser-environment failure. The different total reflects nested test cases that cannot run when module loading fails. These three files are byte-for-byte moves, so the new missing-file errors come from the paths, not changed audio logic.

BalasaitejaG:
Location: server.js — line 11

The old bootstrap caught a failed donation MongoDB connection in a background task and still called app.listen. The new start() awaits connectDatabases before creating or listening to the app, and a rejection propagates to the top-level error handler. A stubbed failure reproduces that neither createApp nor listen is reached. Therefore a database outage at startup now takes down the whole HMI, including login and backend proxying. This is a changed availability policy in a structural refactor.

BalasaitejaG:
Location: server/config/environment.js — line 7

Setting API_PORT=9100 makes loadEnvironment().apiPort equal 9100, but server/services/apiClient.js still constructs http://localhost:9000. createApp does not use dependencies.config for the backend origin; sensor.routes.js and controller/auth.controller.js also hardcode 9000. The new .env.example and both README changes present API_PORT as operational configuration.

Vamshi-Gollapelly (commented):
Went through this one carefully, including the fixes in e23f13b, and I want to say up front the follow-up work here is solid. I checked each of the five things BalasaitejaG flagged against the actual repo, not just the commit message:

The /pages/map guard is back in place and ordered correctly before the static middleware, with a test covering both the blocked and authenticated cases. The tabler-icons import I went and looked at where the file actually lives now, and yeah, the new relative path resolves fine. Same with the spectrogram test paths, the audio_recorder → audio-recorder rename threw me for a second but I confirmed that's genuinely what the file's called now, not a typo. The startup/DB-outage fix has a real regression test showing the server still comes up when Mongo's down, and the API_PORT stuff is threaded through everywhere it needed to be, with tests for each spot.

One small thing I'd want the author to just confirm, not blocking: the fallback path builds a bare new MongoClient(...) when the database connection fails at startup, rather than going through whatever connectDatabases normally does if that function sets other options (auth, TLS, pooling), this bypasses them. Probably fine, just worth a sanity check.

Here's my hesitation on approving though this still shows merge conflicts, and it's 46 commits behind main. Whatever gets resolved during that rebase isn't something I've seen yet, and with 3,339 files touched, it wouldn't take much for a conflict resolution to quietly undo one of these fixes (imagine accepting the wrong side on app.js and losing the checkUserSession line back). I don't want to approve a snapshot that's about to get rewritten underneath it. So, I'd rather hold off and take another look once it's rebased and conflict-free the actual code changes look good to me, this is purely about wanting to review what will really get merged.

PR #1067, opened 2026-09-18
EE/KN: Implementation of Time Warp and Its Augmentation Benchmarking

KCP17:
- **add time warping as an alternative augmentation**: implements `time_warp.py`, a non-masking augmentation operation that stretches/compresses the spectrogram along time instead of deleting content, plus its preset (`config/augmentation/time_warp.yaml`), benchmarked against the existing five masking-based and none-augmented presets on both (unbalanced and balanced) datasets.
- **fix checkpoint-format compatibility**: updates `evaluate_checkpoint()` to handle both the original bare `model.state_dict()` checkpoint format and the newer `{model_state_dict, optimizer_state_dict, ...}` wrapper dict introduced upstream for checkpoint-resume support, so older archived checkpoints keep loading unchanged.
- **update results**: adds the `time_warp` results to `results/README.md` in the new "Addendum" section with the updated final selected configuration, and both `results/*/ceiling40_seed42.json` data files.


Mangon3 (changes requested):
Sorry, my mistake on the incorrect change request

Mangon3 (approved):
Nevermind, LGTM!

PR #1036, opened 2026-09-12
feat(backend): add detection ETags and correlated errors

jayavardhan438-cell:
## Summary
- validate detection rules and map domain/storage failures to stable API errors
- support conditional GET with ETags and `304 Not Modified` responses
- propagate correlation IDs through response headers, error bodies, request state, and logs
- remove the duplicate FastAPI application construction and add regression coverage

## Testing
- `pytest src/tests/unit/backend -q`
- **104 passed**

## Checklist
- A3.2 Detection Rule Error Handling
- B4.1 Conditional GET with ETag Support
- C9.2 Correlation ID Logging Integration
- C16.1 Error-Handling Regression and TD002 Closure

PR #1068, opened 2026-09-18
Add configurable detection rule engine (A3.1)

AshanVimodh:
Introduce app/detection_rules.py: a DetectionRules model plus evaluate_detection(), applied to every incoming detection before it's persisted (app.detections.create_detection). Checks three configurable rules in order - min_confidence, allowed_species, allowed_sensor_ids - each returning a stable machine-readable reason (below_min_confidence, species_not_allowed, sensor_not_allowed) instead of a generic rejection. A rejected detection returns 422 and logs the reason server-side, never the audio payload.

Rule values are read through the C4 Settings model rather than hardcoded in the route: DETECTION_MIN_CONFIDENCE, DETECTION_ALLOWED_SPECIES, DETECTION_ALLOWED_SENSOR_IDS. Settings.Config.parse_env_var is extended so the two list fields accept plain comma-separated values in .env instead of requiring JSON-array syntax. Leaving all three unset preserves today's behaviour exactly (no restrictions).

evaluate_detection() takes a plain object with .confidence/.species/ .sensorId rather than anything HTTP-specific, so A1's MQTT ingestion bridge can call the same function once it exists, instead of duplicating the rule logic for a second ingestion path.

Adds 13 unit tests (accept/reject per rule, the min_confidence boundary, empty allow-lists meaning no restriction, invalid rule values rejected by validation, and env parsing) plus .env.example documentation. Verified live against the real API container: default config behaves identically to pre-A3 (regression-safe), and a configured min_confidence/ allowed_species combination correctly rejects/accepts through the actual POST /detections route.

### DIFF
--- a/src/production/backend/.env.example
+++ b/src/production/backend/.env.example
@@ -15,6 +15,15 @@ MAIL_USERNAME=your_gmail_address_here
 MAIL_PASSWORD=your_gmail_app_password_here
 MAIL_FROM=your_gmail_address_here
 
+# --- Detection rule engine (A3, optional) ---
+# Applied to every incoming detection (app.detection_rules.evaluate_detection)
+# before it's persisted, via app.detections.create_detection. Change these
+# values and restart the API - no route code needs editing. Leaving a value
+# unset/empty means "no restriction" for that rule.
+# DETECTION_MIN_CONFIDENCE=80
+# DETECTION_ALLOWED_SPECIES=Koala,Sus Scrofa
+# DETECTION_ALLOWED_SENSOR_IDS=1,2
+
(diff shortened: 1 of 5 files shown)


Maneesh-Kumar-Maddini (approved):
The configurable detection rules are implemented in the normal detection-create path, including confidence thresholds, species/sensor allowlists, stable rejection reasons, and environment parsing. The 13 focused tests passed in an isolated review run. I found no blocking issue in this scope. Please keep the description clear that engine, MQTT, and bulk ingestion require their own integration. When combining this with #1036, preserve both sets of functionality in app/detection_rules.py.

PR #1009, opened 2026-09-04
EB/MT: Resilience and Scalability changes

Mt-Dkin:
Sprint 2 Resilience Tasks (C1, C3, C6, C18)

PR #1074, opened 2026-09-20
feat(backend): add asynchronous audio processing jobs

nandu1231239:


AshanVimodh (changes requested):
Requesting changes - a few things need addressing before this merges:

**Blocking:**
This branch conflicts with main (25 commits behind), in app/main.py. Please rebase/merge main and resolve.

**Logic issues (not fixed by resolving the conflict - these are all in the new audio_jobs.py files):**
- queue_audio_job does the file write and two Mongo inserts (AudioUploads.insert_one, audio_jobs.create_job) synchronously inside an async def handler, with no asyncio.to_thread. engine.py's event persistence explicitly moves Mongo work off the event loop for this reason (so open WebSocket clients stay responsive) - worth following the same pattern here.
- process_job only catches (OSError, MultiModalPredictionError). Any other exception during prediction propagates uncaught out of the background task, leaving the job stuck in "processing" forever with no error recorded and no way to retry it (retry only works on "failed" jobs).
- If AudioUploads.insert_one succeeds but audio_jobs.create_job fails, the uploaded file gets deleted but the AudioUploads document is left behind - an orphaned metadata record pointing at a file that no longer exists.
- An invalid job_id on POST /audio/jobs/{job_id}/retry returns 409 "Only failed jobs can be retried" instead of 400 "Invalid job ID", since retry_job's _job_id() raises ValueError for both cases and the router's except ValueError maps it to 409 regardless of cause.

**Worth discussing before merge:**
This adds an in-process job mechanism via FastAPI's BackgroundTasks. C1 (#1009/#1048) is building a separate Redis/RQ-backed job queue. If both land, we'd have two different background-job patterns in the codebase - worth agreeing which one is canonical.

PR #1078, opened 2026-09-20
secret loader

Raghav1103-hub:


AshanVimodh (changes requested):
CI is green (lint, build, Docker check), but worth flagging: src/tests/test_secret_loader.py is in a location nothing actually collects. Root pytest.ini's testpaths only lists src/tests/unit/engine, src/tests/integration/engine_backend, src/tests/pipeline, and src/production/engine - backend tests are deliberately excluded there (per TDD_Guide.md, backend needs its own venv with pydantic<2.0) and instead live under src/tests/unit/backend/ (see test_correlation_id.py, test_detection_rules.py, etc.), run explicitly via src/production/backend/.venv/Scripts/python.exe -m pytest <path>.

As placed, this test won't run in either the default pytest invocation or the backend-specific one - the 5 test cases look solid, but they're currently dead. Could you move it to src/tests/unit/backend/test_secret_loader.py and confirm it passes under the backend venv?

Separately, since it doesn't look like anything calls load_secret() yet - is the plan to wire this into app/config.py's Settings model, or is it meant to replace/extend the C5 *_FILE mechanism there? Right now it's a second, disconnected secret-handling path, and it'd be good to reconcile which one is canonical before this merges.

PR #1080, opened 2026-09-22
Centralise Cloudflare R2 credentials in backend configuration

thomas-pham101:
## Summary

Centralises Cloudflare R2 configuration within the Backend `Settings`
architecture instead of reading R2 credentials directly from environment
variables inside the storage module.

## Changes

- Added R2 configuration fields to `app/config.py`
  - `R2_ACCOUNT_ID`
  - `R2_ACCESS_KEY_ID`
  - `R2_SECRET_ACCESS_KEY`
  - `R2_BUCKET_NAME`
  - `R2_ENDPOINT_URL`
- Added `R2_DATASET_PREFIX` to central settings to remove remaining direct
  environment access from the R2 module.
- Added R2 placeholders to the Backend `.env.example`.
- Added secret-file support for R2 access credentials.
- Refactored `R2Config` to consume validated Backend settings rather than
  `os.environ`.
- Updated the R2 demo and documentation to use the central Backend configuration.
- Updated R2 and Backend configuration tests.

## Validation

- Backend configuration tests: **16 passed**
- Cloudflare R2 unit tests: **9 passed**
- R2 settings mapping smoke test: **PASS**
- Missing R2 configuration validation: **PASS**
- Backend startup without R2 credentials: **PASS**
- R2 demo import using Backend configuration: **PASS**
- Confirmed no remaining `os.getenv()` / `os.environ` reads in the R2 Python module
- Confirmed no remaining `R2Config.from_env()` callers
- `git diff --check`: **PASS**

## Scope

This PR only centralises R2 configuration.

Routing audio uploads into Cloudflare R2 is intentionally left for the
separate follow-up task.


AshanVimodh (approved):
Approved since nothing is code breaking. One non-blocking design question: endpoint_url used to be computed from account_id automatically, and this PR makes it a separate required field (R2_ENDPOINT_URL) instead - was there a reason to make it independently configurable rather than derived? If not, worth simplifying back; if yes (e.g. testing against a mock endpoint), a comment in .env.example would help.

thomas-pham101:
Thanks for the feedback. The main reason I kept R2_ENDPOINT_URL independently configurable was to allow alternative or mock S3-compatible endpoints for testing and local development. I agree the reason isn’t obvious from the current config, so adding a short note in .env.example would make sense.

PR #1065, opened 2026-09-17
Backend Sprint 2: MQTT, audio, bulk ingestion and IoT integration

seudhipa404:
## Sprint 2 Backend Integration

Completed assigned Backend Sprint 2 tasks:

- A1.1–A1.2 MQTT subscriber and persistence integration
- A2.1 Audio clip storage/retrieval and detection linking
- A5.1 Bulk detection ingestion API
- A6.1 IoT-to-Backend end-to-end integration verification

## Verification

- Backend connects and subscribes to the MQTT detection topic
- Valid MQTT detections are validated and persisted to MongoDB
- Invalid MQTT payloads are rejected and not persisted
- Audio files can be uploaded and retrieved with matching SHA256 hashes
- Detection events can reference stored audio upload IDs
- Bulk ingestion supports valid batches and partial rejection
- IoT/MQTT event verified through Backend → MongoDB → HMI latest-events retrieval
- Final regression suite: 21 tests passed

## Notes

- Local detection budget was configured only for integration testing.
- Existing Starlette python_multipart deprecation warning remains unrelated to this work.

PR #1082, opened 2026-09-22
Connect audio upload ingestion to Cloudflare R2

thomas-pham101:
## Summary

Routes Backend audio uploads through Cloudflare R2 instead of writing uploaded
audio files to local temporary disk.

This PR is stacked on the R2 configuration centralisation PR and contains only
the audio-ingestion integration changes.

## Changes

- Route `POST /api/audio/upload` directly to Cloudflare R2
- Remove local upload-directory persistence
- Return a persistent `storage_key` in the API response
- Persist the R2 storage key in `AudioUploads` metadata
- Add clean R2 failure handling:
  - `503` when R2 is unavailable
  - `502` when the upstream upload fails
- Add best-effort R2 cleanup if Mongo metadata persistence fails
- Add `delete_object()` support to the shared R2 storage client
- Add Backend R2 storage adapter
- Add `boto3` to Backend runtime dependencies
- Package the shared R2 module into the Backend Docker image
- Update Docker Compose build contexts and R2 environment examples
- Add automated audio-upload/R2 integration tests

## Validation

- Audio upload + configuration regression: **23 passed**
- Shared R2 storage unit tests: **10 passed**
- Docker Compose configuration: **PASS**
- Docker Compose test configuration: **PASS**
- Backend runtime imports (`boto3`, `cloudflare_r2`, R2 service): **PASS**
- `git diff --check`: **PASS**

### Covered upload scenarios

- Successful upload -> R2 + Mongo metadata + `storage_key`
- R2 unavailable -> `503`
- R2 upload failure -> `502`
- Mongo persistence failure -> `500` + R2 cleanup
- Invalid extension -> `400`
- Empty upload -> `400`
- Oversized upload -> `413`

## Dependency

This PR is currently based on the R2 backend configuration PR:

`EB/TP/R2-Centralised-Backend-Config`

Once that PR is merged, this branch can be rebased/retargeted onto `main`.


AshanVimodh:
@thomas-pham101 Since #1080 merged, this PR's base is still `EB/TP/R2-Centralised-Backend-Config` instead of `main`. Merging as-is would merge the R2 audio-upload changes into that feature branch, not into `main` - the PR would show as merged but the actual code wouldn't land where it needs to.

Could you retarget this PR's base to `main` (Edit next to the base/compare branches near the top of the page), or delete `EB/TP/R2-Centralised-Backend-Config` now that it's merged so GitHub retargets it automatically? Once that's done we can review/merge this properly.

thomas-pham101:
@AshanVimodh  Thanks for catching this. I’ve retargeted #1082 to main now that #1080 has merged. I’ve verified the PR diff is limited to the audio-upload R2 integration changes, and the validation checks are passing. It should now be ready for review.

AshanVimodh:
@thomas-pham101 One thing to sort out before this merges: [#1065](https://github.com/DataBytes-Organisation/Project-Echo/pull/1065) (open, approved, adds GET /audio/{upload_id}) touches this same file and will collide with this PR's storage change.

1065 appends a GET /audio/{upload_id} endpoint that reads audio back from local disk via UPLOAD_DIR and meta.get("filename"). This PR removes UPLOAD_DIR entirely and switches uploads to R2 with a storage_key, and doesn't add a replacement retrieval endpoint.

I tested the merge directly: whichever of these two lands second, resolving the conflict won't be enough on its own.
- If 1065 merges first and this one lands after: UPLOAD_DIR's removal is in a part of the file 1065 never touches, so it merges cleanly with no conflict marker - but 1065's GET endpoint is left calling os.path.join(UPLOAD_DIR, filename) with UPLOAD_DIR no longer defined anywhere. That's a NameError the first time anyone hits the endpoint, and it won't get caught by CI - the lint step only checks E9,F63,F7,F82 (not undefined-name), and the Docker build check doesn't execute the endpoint.
- If this PR merges first and 1065 lands after: git does flag a real conflict in this file (the GET function sits right after the return statement both PRs touch), so it's visible - but resolving it still means someone has to actually rewrite the GET endpoint to pull from R2 via storage_key rather than local disk, not just pick a side.

Could you add a GET (or equivalent retrieval) path in this PR that serves the file back from R2 by storage_key? That way 1065's endpoint becomes redundant/replaceable regardless of merge order, and whoever resolves the eventual conflict has a clear R2-based version to reconcile against instead of two incompatible storage assumptions.

thomas-pham101:
@AshanVimodh Thanks for catching the compatibility issue with #1065. I’ve now updated #1082 so GET /api/audio/{upload_id} retrieves audio directly from Cloudflare R2 using the persisted storage_key, with no dependency on UPLOAD_DIR.

I also added handling for unavailable storage, missing R2 objects, and download failures.

Validation:

+) Audio upload/retrieval tests: 14 passed
+) Shared Cloudflare R2 unit tests: 10 passed
+) git diff --check: PASS
+) No UPLOAD_DIR dependency remains in the audio router

The latest changes have been pushed and the branch is synced with main. Could you please re-review #1082 when you have a chance? Thanks.

thomas-pham101:
Hi @AshanVimodh , I’ve resolved the conflict with the newly merged #1065.

The audio-linking tests from #1065 have been updated to use the R2-backed retrieval flow rather than the previous local UPLOAD_DIR implementation.

Validation after resolving the conflict:

+) Audio upload + linking tests: 19 passed
+) Combined relevant regression suite: 37 passed
+) git diff --check: PASS
+) No UPLOAD_DIR dependency remains in the audio router

GET /api/audio/{upload_id} continues to retrieve the object from R2 using the persisted storage_key.

Could you please re-review #1082 when you have a chance? Thanks.

